PR #6, opened 2016-12-16
Update domino.py

johnjoo1:
urllib2 -> urllib.request to make Python 3 compatible


domino/domino.py
```diff
@@ -1,5 +1,8 @@
 from .routes import _Routes
-import urllib2
+try:
+    import urllib2
+except ImportError:
+    import urllib.request as urllib2
```
farmdawgnation: It seems like an anti-pattern to catch an error to determine what we should import. Could we use the [six](https://pypi.python.org/pypi/six) library to detect Python3 and use that to determine what to import?

PR #8, opened 2017-01-18
Add upload file functionality

marks:
@johnjoo1 - could you take a look and confirm this works? There's an example in the `examples/` dir.


examples/upload_file.py
```diff
@@ -0,0 +1,16 @@
...
+
+# connect to domino; be sure to have these environment variables set
+#  (runs inside a Domino executor automatically set these for you)
+domino = Domino("marks/test", api_key=os.environ['DOMINO_USER_API_KEY'], host=os.environ['DOMINO_API_HOST'])
+
+f = open('numbers.csv', 'rb')
+r = domino.files_upload("/a_new_folder/a_new_file.csv", f)
+
+if r.status_code == 201:
+	print "Upload successful"
```
johnjoo1: should make this compatible with Python 3.

examples/upload_file.py
```diff
@@ -0,0 +1,16 @@
...
+else:
+	print "!! Upload failed"
```
johnjoo1: same here

marks:
@johnjoo1 can you take a look at this and agree if it can be merged? It technically will not work until we address PR #6 for py3 compat but I want to do one thing at a time

marks:
Spoke with John... merging

Issue #36, opened 2018-04-28
Add (temporary, beta) model manager APIs 


marks:
https://dominodatalab.github.io/api-docs/#models-domino-2.5+

Reminder, these are temporary/beta as noted in the API docs:

> These are beta endpoints. They will not have long term support beyond Q3 2018.
> 
> These endpoints are only available in Domino 2.5+. These APIs allow you to interact with your project’s models.

- [x] List environments for deployment
- [x] List models for project
- [x] Publish model
- [x] List model versions
- [x] Publish model version

marks:
Done and merged

PR #37, opened 2018-04-28
Add model manager (2.5+) APIs

marks:
See issue #36 for details as well

### DIFF
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -156,12 +156,12 @@ class Domino:
             "ignoreRepoState": False
         }
         response = requests.post(url, auth=('', self._api_key), json=request)
-        
+
         if response.status_code == 400:
             raise Warning("Run ID:" + runId + " not found.")
         else:
-            return response            
-    
+            return response
+
     def runs_status(self, runId):
         url = self._routes.runs_status(runId)
         return self._get(url)
@@ -198,7 +198,7 @@ class Domino:
 
     def fork_project(self, target_name):
         url = self._routes.fork_project()
-        request = { "overrideProjectName" : target_name}
+        request = {"overrideProjectName": target_name}
         response = requests.post(url, auth=('', self._api_key), data=request)
         return response.status_code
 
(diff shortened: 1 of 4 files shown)


marks:
@johnjoo1 would you mind giving this a once over?

marks:
@macrogers you have time to take a look and/or test this even?

domino/domino.py
```diff
@@ -264,21 +275,75 @@ def collaborators_add(self, usernameOrEmail, message=""):
...
+        response = requests.post(url, auth=('', self._api_key), json=request)
+        return response.json()
+
+    def model_versions_get(self, model_id):
+        self.requires_at_least("2.4.0")
+        url = self._routes.model_versions_get(model_id)
+        return self._get(url)
+
+    def model_version_publish(self, model_id, file, function, environment_id,
+                              name, description="A great model",
```
johnjoo1: Description default seems a bit gimicky and unnecessary. Feels like it would be annoying to set that as blank all the time.

domino/routes.py
```diff
@@ -4,20 +4,30 @@ def __init__(self, host, owner_username, project_name):
...
     def _build_project_url(self):
         return self.host + '/v1/projects/' + \
             self._owner_username + '/' + self._project_name
 
     def _build_project_url_private_api(self):
-        return self.host + '/u/' + self._owner_username + '/' + self._project_name
+        return self.host + '/u/' + self._owner_username + \
+            '/' + self._project_name
+
+    def _build_old_project_url(self):
```
johnjoo1: It doesn't seem like we still need this. Should we just yank this out?

domino/domino.py
```diff
@@ -264,21 +275,75 @@ def collaborators_add(self, usernameOrEmail, message=""):
...
+            "description": description,
+            "projectId": self._project_id,
+            "file": file,
+            "function": function,
+            "excludeFiles": files_to_exclude,
+            "environmentId": environment_id
+        }
+
+        response = requests.post(url, auth=('', self._api_key), json=request)
+        return response.json()
```
johnjoo1: should we just return the published model id? Do you see the user using the other fields in this?

marks: Talked IRL.. going to return everything

domino/routes.py
marks: Talked IRL.. used in some places until D3.0 APIs

domino/domino.py
marks: Yeah - #weekendCoding. Will set to blank

PR #40, opened 2018-05-25
Replace json decode errors with status codes

Spacerat:
Currently, all commands which unwrap responses using `.json()` fail with JSON decode errors if the request failed.

This diff is a quick fix using `response.raise_for_status()` to raise an exception with the error code instead. I also have a little wrapper which plucks out the error message in cases where it is not HTML.

### DIFF
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -5,6 +5,7 @@ try:
 except ImportError:
     import urllib.request as urllib2
 
+from future.utils import raise_from
 import os
 import logging
 import requests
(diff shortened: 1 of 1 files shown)


domino/domino.py
```diff
@@ -75,6 +76,7 @@ def runs_start(self, command, isDirect=False, commitId=None, title=None,
         }
 
         response = requests.post(url, auth=('', self._api_key), json=request)
+        raise_status_with_body(response)
```
abhijeet2096, line 79: Please create _post wrapper method (like _get) and consolidate this (raise_status_with_body) in it

Spacerat:
Hi! I’m no longer using Domino, so this PR is in your hands - I won’t be updating it

abhijeet2096:
Changes from this P.R have been incorporated as part of these P.Rs

https://github.com/dominodatalab/python-domino/pull/76
https://github.com/dominodatalab/python-domino/pull/77

PR #42, opened 2018-10-17
Support Ticket 45332 - TypeError in Python Domino Package

imarchenko:
Fixed _project_id to fix response binary->string conversion


imarchenko:
@marks This pull request contains the change for making https://github.com/dominodatalab/python-domino/blob/master/examples/models_and_environments.py work with Python 2 and 3. The only change needed in models_and_environments.py, to make it work with Python 3, is line 10:

global_environments = filter(...) -> global_environments = list(filter(...))

marks:
@imarchenko cool - can you please make the change to the example file too and merge this?

PR #44, opened 2019-02-03
Fix version confusion

marks:
by consolidating where we keep the version information to just __init.py__ as suggested by https://packaging.python.org/guides/single-sourcing-package-version/

(relevant zendesk ticket #46637)


ddl-alexpanin:
Closing as versioning was already addressed.

PR #46, opened 2019-03-14
DOM-12518: update python client to account for new hardware tier argument for app publishing

andrewziegler:
Updates `app_publish` to accept an optional `hardwareTierId` parameter which is introduced to the api in https://github.com/cerebrotech/domino/pull/14351.

See https://dominodatalab.atlassian.net/browse/DOM-12518 for more details.

### DIFF
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -285,10 +285,7 @@ class Domino:
         if unpublishRunningApps is True:
             self.app_unpublish()
         url = self._routes.app_publish()
-        if hardwareTierId is None:
-            request = {"language": "App"}
-        else:
-            request = {"language": "App", "hardwareTierId": hardwareTierId}
+        request = {"language": "App", "hardwareTierId": hardwareTierId}
         response = requests.post(url, auth=('', self._api_key), json=request)
         return response
 
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -281,11 +281,14 @@ class Domino:
             return disposition
 
     # App functions
-    def app_publish(self, unpublishRunningApps=True):
+    def app_publish(self, unpublishRunningApps=True, hardwareTierId=None):
         if unpublishRunningApps is True:
             self.app_unpublish()
         url = self._routes.app_publish()
-        request = {"language": "App"}
+        if hardwareTierId is None:
+            request = {"language": "App"}
+        else:
+            request = {"language": "App", "hardwareTierId": hardwareTierId}
         response = requests.post(url, auth=('', self._api_key), json=request)
         return response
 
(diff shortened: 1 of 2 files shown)


domino/domino.py
```diff
@@ -281,11 +281,14 @@ def collaborators_add(self, usernameOrEmail, message=""):
...
-    def app_publish(self, unpublishRunningApps=True):
+    def app_publish(self, unpublishRunningApps=True, hardwareTierId=None):
         if unpublishRunningApps is True:
             self.app_unpublish()
         url = self._routes.app_publish()
-        request = {"language": "App"}
+        if hardwareTierId is None:
+            request = {"language": "App"}
+        else:
+            request = {"language": "App", "hardwareTierId": hardwareTierId}
```
robinkraft: Is this if/else unnecessary? Doesn't the API behave the same way either way, since `hardwareTierId` is implicitly `None` if you don't pass it in?

andrewziegler: from looking elsewhere in this file I think you are correct. I think I can just write:
```
request = {"language": "App", "hardwareTierId": hardwareTierId}
```

thanks

andrewziegler: updated.

domino/domino.py
```diff
@@ -281,11 +281,11 @@ def collaborators_add(self, usernameOrEmail, message=""):
             return disposition
 
     # App functions
-    def app_publish(self, unpublishRunningApps=True):
+    def app_publish(self, unpublishRunningApps=True, hardwareTierId=None):
         if unpublishRunningApps is True:
             self.app_unpublish()
         url = self._routes.app_publish()
-        request = {"language": "App"}
+        request = {"language": "App", "hardwareTierId": hardwareTierId}
```
fernandoacorreia: @robinkraft make sure this still works for old code that doesn't pass a hardware tier ID

andrewziegler:
@robinkraft DOM-12518 is marked as done ... can we merge this? Who has permissions to merge this?

fernandoacorreia:
From my part it can be merged after @robinkraft validates that it's working in all cases (with and without HT).

Please ask @chrismyang about how to get the updated version on pip (we may need a separate ticket to track that if it's complicated).

robinkraft:
Tested successfully. #shipit.

domino/domino.py
robinkraft: All good.

robinkraft:
Btw @fernandoacorreia this package is not available on pypi. According to the readme, you have to install it this way:
`pip install https://github.com/dominodatalab/python-domino/archive/master.zip`

Or with `requirements.txt`: `-f git+git://github.com/dominodatalab/python-domino.git`

PR #47, opened 2019-06-25
Fixed publish_ui() to use updated URL in Domino

imarchenko:
The path /u/<user_name>/<project_name>/endpoints returns a 404 code now in newer versions of Domino. It seems we have internally updated the path to be /u/<user_name>/<project_name>/endpoints/modelManager, so publish_ui() function was updated to reflect the change. Without this fix, anything that relies on publish_ui() breaks. This particular code breaks:
https://github.com/dominodatalab/python-domino/blob/master/domino/domino.py#L380-L390

### DIFF
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -381,6 +381,9 @@ class Domino:
     @property
     def _project_id(self):
         url = self._routes.publish_ui()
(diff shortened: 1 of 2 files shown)


johnjoo1:
lgtm

PR #49, opened 2019-12-31
Added get datasets support.

katieshakman:
Added a function datasets_list and corresponding route.  Also added get_datasets.py in the examples folder.


katieshakman:
Cleaning up a comment.  Will open a new PR.

PR #51, opened 2020-01-29
Airflow DominoOperator

wseaton:
Is there any interest in taking work that I've done to make a `DominoOperator` for airflow and bundling it with this project as an optional extra package? 

As a bonus side effect it could help greatly clean up the documentation in this section: https://docs.dominodatalab.com/en/4.1/reference/runs/advanced/Using_Apache_Airflow_with_Domino.html

The `PythonOperator` is recommended, but as an end user there are some pitfalls...

We use this internally for scheduling runs and it has some extra features to work around kinks of the API, mainly the fact that setup for runs isn't logged by default and functionality to help mitigate the race condition in 3.6 when scheduling multiple runs in parallel.

I would consider adding it to Airflow's operator contrib repository proper but I don't think there would be much community interest. Would be happy to submit the code to you guys here and let representatives from domino weigh-in and/or go through the Apache JIRA feature request process yourselves.

Some notes since this is WIP:
- Type hints would need to be made backwards compatible. (Unless we make this a Python3 only feature)
- Exception handling needs to be added if the associated extra dependencies aren't installed
- Documentation needs to be added, probably for just extra kwargs that don't follow the function signature of their wrapped counterparts

Thanks!

### DIFF
--- a/domino/airflow/__init__.py
+++ b/domino/airflow/__init__.py
@@ -1,0 +1,4 @@
+try:
+    from domino.airflow._operator import DominoOperator
+except SyntaxError:
+    raise ImportError("Use of the Airflow DominoOperator requires typing (Python 3.5+).")
--- a/domino/airflow/_operator.py
+++ b/domino/airflow/_operator.py
@@ -1,0 +1,121 @@
+import time
+from typing import List, Optional, Any
+
+from bs4 import BeautifulSoup
+
+from airflow.models import BaseOperator
+from airflow.utils.decorators import apply_defaults
+from domino import Domino
+
+
+class DominoOperator(BaseOperator):
+    """
+    Operator for interacting with Domino Data Lab
+    via the python-domino client library w/ some
+    additional functionality baked in. Follows the 
+    same run signature as domino.runs_start with
+    some extra arguments.
+    
+    Host and API key arguments are optional and can be
+    discovered via environment variable, as per the domino
+    python client.
+
+
+    Notes
+    ------
+    When combining `isDirect=True` with a command, you
+    need to pass in the entire command as a single string
+    in the command array. 
+    """
+
+    template_fields = ("command", "title")
+    ui_color = "#5188c7"
+
+    @apply_defaults
+    def __init__(
+        self,
+        project: str,
+        command: List[str],
+        host: Optional[str] = None,
+        api_key: Optional[str] = None,
+        domino_token_file: Optional[str] = None, 
+        isDirect: bool = None,
+        commitId: Optional[str] = None,
+        title: Optional[str] = None,
+        tier: Optional[str] = None,
+        publishApiEndpoint: Optional[bool] = None,
+        poll_freq: Optional[int] = 5,
+        max_poll_time: Optional[int] = 6000,
+        startup_delay: Optional[int] = 10,
+        include_setup_log: Optional[bool] = True,
+        *args,
+        **kwargs
+    ):
+        super(DominoOperator, self).__init__(*args, **kwargs)
+
+        self.log.info("Initializing Client...")
+        
+        self.project = project
+        self._api_key = api_key
+        self._host = host
+        self._domino_token_file = domino_token_file
+        self.command = command
+        self.is_direct = isDirect
+        self.commit_id = commitId
+        self.title = title
+        self.tier = tier
+        self.publish_api_endpoint = publishApiEndpoint
+        self.poll_freq = poll_freq
+        self.max_poll_time = max_poll_time
+        self.startup_delay = startup_delay
+        self.include_setup_log = include_setup_log
+
+        self.client = None
+        self.run_id = None
+
+    def execute(self, context: Any) -> dict:
+
+        self.client = Domino(self.project, self._api_key, self._host, self._domino_token_file)
+        self.log.info("Client Initialized for project: %s", self.project)
+
+        self.log.info("Starting run.")
+        if self.startup_delay:
+            time.sleep(self.startup_delay)
+
+        if self.is_direct:
+            if len(self.command) > 1:
+                raise ValueError(
+                    "Domino API will not accept a "
+                    "multipart command string this long if is_direct=True"
+                )
+
+        run_response = self.client.runs_start_blocking(
+            command=self.command,
+            isDirect=self.is_direct,
+            commitId=self.commit_id,
+            title=self.title,
+            tier=self.tier,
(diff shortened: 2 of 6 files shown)


wseaton:
@ddl-alexpanin Is the domino team interested in this PR if the changes I've listed above are made and API is updated to 4.2? Our internal deployment will be updated soon so I'll have a chance to test.

ddl-alexpanin:
> @ddl-alexpanin Is the domino team interested in this PR if the changes I've listed above are made and API is updated to 4.2? Our internal deployment will be updated soon so I'll have a chance to test.

@wseaton thanks for working on this pull request! We are definitely interested.

wseaton:
@ddl-alexpanin I was able to test today and the `DominoOperator` works just fine on both ddl 4.2.2 and 3.6 with either python client versions (they appear to all be cross compatible as long as you use an API_KEY for auth). 

I'm hesitant to remove the type hints since it is airflow convention now for operators to include them. Additionally, I'm not sure targeting python 2.7 support is worthwhile if the airflow community is actively moving away from it being supported for Airflow 2.0, so my easy solution is just to throw an error if a user attempts to import the package from an unsupported version.

Was also able to add domino token authfile support easily enough.

domino/airflow/__init__.py
```diff
@@ -0,0 +1,4 @@
+try:
+    from domino.airflow._operator import DominoOperator
+except SyntaxError:
```
wseaton, line 3: If we don't do this python throws a compile time exception (due to `SyntaxError` because typing is a language feature) which means the entire module can't be loaded.

domino/airflow/_operator.py
```diff
@@ -0,0 +1,165 @@
...
+            commitId=self.commmit_id,
+            title=self.title,
+            tier=self.tier,
+            publishApiEndpoint=self.publish_api_endpoint,
+        )
+
+        self.run_id = run_response["runId"]
+
+        poll_start = time.time()
+        while True:
```
abhijeet2096, line 117: Hey, Can we use `run_start_blocking` method instead of using `runs_start`, I see the only difference is the logs which we are passing into airflow. Can we change it something like below

```
run_response = self.client.runs_start_blocking(<arguments-here>)
log = self.get_run_log()
# spool out and replay entire log
      for line in log.splitlines():
            if line:
            # using bs4 to strip the HTML tags
                  text = BeautifulSoup(line).text
                  if "text-danger" in line:
                       self.log.warning(text)
                  else:
                        self.log.info(text)
return run_response
```

domino/airflow/_operator.py
```diff
@@ -0,0 +1,165 @@
...
+
+    @apply_defaults
+    def __init__(
+        self,
+        project: str,
+        command: List[str],
+        host: Optional[str] = None,
+        api_key: Optional[str] = None,
+        domino_token_file: Optional[str] = None, 
+        isDirect: bool = None,
```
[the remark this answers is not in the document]
wseaton: is there a reason for this? while it may match the API convention it is not PEP8 friendly.

domino/airflow/_operator.py
wseaton: Ah yes, since logs are only made available after a run completes and we can't pull them async this makes sense. I think originally I had planned to do that but realized it wasn't possible. I'll change it.

domino/airflow/_operator.py
```diff
@@ -0,0 +1,165 @@
...
+        self.publish_api_endpoint = publishApiEndpoint
+        self.poll_freq = poll_freq
+        self.max_poll_time = max_poll_time
+        self.startup_delay = startup_delay
+        self.include_setup_log = include_setup_log
+
+        self.client = None
+        self.run_id = None
+
+    def get_run_log(self) -> str:
```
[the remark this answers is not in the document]
wseaton: Ok, I'll drop the type hint to preserve backwards compatibility.

domino/airflow/_operator.py
```diff
@@ -0,0 +1,165 @@
...
+
+        self.log.info("Initializing Client...")
+        
+        self.project = project
+        self._api_key = api_key
+        self._host = host
+        self._domino_token_file = domino_token_file
+        self.command = command
+        self.is_direct = isDirect
+        self.commmit_id = commitId
```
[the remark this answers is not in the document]
wseaton: Good catch!

wseaton:
My apologies, please allow me to clean up the commit history, did a bad rebase attempting to pick up changes from `1.0.0`.

wseaton:
Okay, squashed changes, ready to keep reviewing.

wseaton:
> Hey @wseaton,
> 
> Thanks for the P.R. It looks good, have some minor comment. About python 2 support, I don't think it is required. It will be really helpful if you could add some documentation (in the readme) and a example of using the same in `examples` folder

Python 2 installation doesn't work for `1.0.0` anyways due to the use of f-strings in the `setup.py`, so this is a non-issue.

```
    ERROR: Command errored out with exit status 1:
     command: /home/weaton/.pyenv/versions/2.7/bin/python2.7 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-req-build-cIXnG5/setup.py'"'"'; __file__='"'"'/tmp/pip-req-build-cIXnG5/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-req-build-cIXnG5/pip-egg-info
         cwd: /tmp/pip-req-build-cIXnG5/
    Complete output (6 lines):
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-cIXnG5/setup.py", line 9
        f = open(f"{PACKAGE_NAME}/_version.py")
                                             ^
    SyntaxError: invalid syntax
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

examples/example_airflow_dag.py
```diff
@@ -0,0 +1,34 @@
+from datetime import datetime, timedelta
+from airflow import DAG
+from airflow.models import Variable
+
+from domino.airflow import DominoOperator
+
+api_key = Variable.get("DOMINO_API_KEY")
```
abhijeet2096, line 7: Minor: Can we rename it to `DOMINO_USER_API_KEY`

PR #58, opened 2020-06-25
[DOM-21559] Add `retry` param in `runs_start_blocking`

abhijeet2096:
### What problem does this P.R solve?
Currently in `runs_start_blocking` method, if Http request fails because of transient http errors, the method fails. 

### Solution
This P.R adds a `retry` param, which will retry the request incase of transient http errors

### Link to Jira
https://dominodatalab.atlassian.net/browse/DOM-21559

image
Case Method fails after Max retries
image
Case Method succeeds after 2 failed request.

### DIFF
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -137,6 +137,7 @@ class Domino:
         while True:
             try:
                 run_info = self.get_run_info(run_id)
+                current_retry_count = 0
             except requests.exceptions.RequestException as e:
                 current_retry_count += 1
                 self._logger.warn(f'Failed to get run info for runId: {run_id} : {e}')
@@ -144,6 +145,7 @@ class Domino:
                     raise Exception(f'Cannot get run info, max retry {retry_count} exceeded') from None
                 else:
                     self._logger.info(f'Retrying ({current_retry_count}/{retry_count}) getting run info ...')
+                    time.sleep(poll_freq)
                     continue
 
             elapsed_time = time.time() - poll_start
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -82,7 +82,7 @@ class Domino:
 
     def runs_start_blocking(self, command, isDirect=False, commitId=None,
                             title=None, tier=None, publishApiEndpoint=None,
-                            poll_freq=5, max_poll_time=6000):
+                            poll_freq=5, max_poll_time=6000, retry_count=5):
         """
         Run a tasks that runs in a blocking loop that periodically checks to
         see if the task is done.  If the task errors an exception is raised.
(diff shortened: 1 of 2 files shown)


domino/domino.py
```diff
@@ -122,14 +122,30 @@ def runs_start_blocking(self, command, isDirect=False, commitId=None,
                         Maximum number of seconds to wait for a task to
                         complete. If this threshold is exceeded, an exception
                         is raised.
+
+        retry_count : int (Optional)
+                        Maximum number of retry to do while polling
+                        (in-case of transient httpd errors). If this
```
abhijeet2096, line 128: httpd -> http

domino/domino.py
```diff
@@ -122,14 +122,30 @@ def runs_start_blocking(self, command, isDirect=False, commitId=None,
...
         run_response = self.runs_start(command, isDirect, commitId, title,
                                        tier, publishApiEndpoint)
         run_id = run_response['runId']
 
         poll_start = time.time()
+        current_retry_count = 0
         while True:
-            run_info = self.get_run_info(run_id)
+            try:
+                run_info = self.get_run_info(run_id)
```
mpkouznetsov, line 139: Do we want to reset the `current_retry_count` to 0 here? I think this would make more sense to the end user. I think the behavior we want is a successful retry resets the count rather than we have the right to up to `retry_count` non-consecutive failures.

domino/domino.py
```diff
@@ -122,14 +122,30 @@ def runs_start_blocking(self, command, isDirect=False, commitId=None,
...
+            except requests.exceptions.RequestException as e:
+                current_retry_count += 1
+                self._logger.warn(f'Failed to get run info for runId: {run_id} : {e}')
+                if current_retry_count > retry_count:
+                    raise Exception(f'Cannot get run info, max retry {retry_count} exceeded') from None
+                else:
+                    self._logger.info(f'Retrying ({current_retry_count}/{retry_count}) getting run info ...')
+                    continue
```
mpkouznetsov, line 147: Do we want to go to sleep before retrying?

domino/domino.py
abhijeet2096: Yes, it makes more sense.

domino/domino.py
abhijeet2096: Will add a sleep about `poll_freq ` here

abhijeet2096:
Hey @mpkouznetsov , We have a similar functionality in Domino CLI ((https://github.com/cerebrotech/domino/blob/9203e2523a6578c1f9d4ebedfb26221685735cc4/client/src/main/scala/domino/client/command/RunStatusCommand.scala#L31)). There we are not doing exponential backoff. Since, we already have a `poll_freq` param, I think we can have linear count for retry (Not necessarily consecutive, will address this comment in the P.R)

PR #59, opened 2020-06-30
[DOM-21654] Update `python-domino` library to use auth tokens

abhijeet2096:
### What Problem does this P.R solve?
Domino 4.3 is introducing auth-tokens in Domino Workspaces and Jobs. This P.R adds support for `Domino Auth Token` (Access Token) to authenticate requests in `python-domino`.

### What is the solution?
This P.R introduces following classes 
1. `_HttpRequestManager` class for making requests (Via python's requests module). This class takes `Auth` instance to initialise and authenticate requests. 
2. `BearerAuth` class for authenticating requests via BearerToken (AccessToken).

`python-domino` class can be initialised now with bearer token as follows:
*  Use `DOMINO_TOKEN_FILE` environment variable
* (Or use) Pass domino_token_file constructor variable for `Domino` 

### Additional Notes
Currently the priority for using authenticating requests is as follows:

Check for Access Token -> if not available Check for API_token.

Incase both of these are available, `prefer_domino_token_file` class constructor variable can be used

### Modified blob_get(key) method
In this P.R I have modified the return type of `blob_get` method

from <class 'http.client.HTTPResponse'>  ->  `<class 'urllib3.response.HTTPResponse'>` . 

This change came because of using `requests` library (Which is used in all other places in the library) in place of urllib2.

### Link to the Ticket
https://dominodatalab.atlassian.net/browse/DOM-21654

### Additional Links
https://requests.readthedocs.io/en/latest/user/quickstart/#raw-response-content

### ToDo
- [x] Update Readme

- [x] Update Support docs


domino/domino.py
```diff
@@ -420,27 +419,20 @@ def model_version_publish(self, model_id, file, function, environment_id,
...
-        handler = urllib2.HTTPBasicAuthHandler(password_mgr)
-        opener = urllib2.build_opener(handler)
-        return opener.open(url)
+        return self.request_manager.put(url, data=file)
 
     def _validate_blob_key(self, key):
         regex = re.compile("^\\w{40,40}$")
         if not regex.match(key):
-            raise Exception(("Blob key should be 40 alphanumeric charachters. "
+            raise Exception(("Blob key should be 40 alphanumeric characters. "
```
abhijeet2096, line 435: Fixed a typo here.

domino/domino.py
```diff
@@ -1,41 +1,36 @@
...
 import requests
+from requests.auth import HTTPBasicAuth
 import time
 import pprint
 import re
 
 
 class Domino:
-    def __init__(self, project, api_key=None, host=None):
+    def __init__(self, project, api_key=None, host=None, domino_token_file=None, prefer_domino_token_file=True):
```
mpkouznetsov, line 21: I am questioning the necessity of giving the user this extra flexibility.

PR #60, opened 2020-07-16
[DOM-21812] Fix project create method

abhijeet2096:
This P.R fixes the Project Create method.

### Link to Jira
https://dominodatalab.atlassian.net/browse/DOM-21812

### DIFF
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -268,19 +268,17 @@ class Domino:
         url = self._routes.deployment_version()
         return self._get(url)
 
-    def project_create(self, owner_username, project_name):
-        self.requires_at_least("1.53.0.0")
+    def project_create(self, project_name, owner_username=None):
         url = self._routes.project_create()
-        request = {
-            'owner': owner_username,
-            'name': project_name
+        data = {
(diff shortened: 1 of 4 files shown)


domino/domino.py
```diff
@@ -268,19 +268,17 @@ def deployment_version(self):
...
+            'projectName': project_name,
+            'ownerOverrideUsername': owner_username
         }
-        response = self.request_manager.post(url, data=request, allow_redirects=False)
-        disposition = parse_play_flash_cookie(response)
-        if disposition.get("error"):
-            raise Exception(disposition.get("message"))
-        else:
-            return disposition
+        headers = {
```
ddl-alexpanin, line 277: Can you explain why is this required here and not in other cases?

abhijeet2096: Hey Alex, I checked our server code, we are applying these CSRF filters on `POST` request with contentType `"application/x-www-form-urlencoded", "multipart/form-data", "text/plain"`. This headers needs to be added in two more places in `python-domino` (fork_project and collaborators_add). I am thinking to do that in a separate P.R

PR #64, opened 2020-08-14
DominoOperator logging fix

### LINKED ISSUE #63: clean up DominoOperator logging
When switching from `runs_start` to `runs_start_blocking`  in #51 one unintended consequence is that because `runs_start_blocking` logs on stdout on it's own internally to the Domino client, the `DominoOperator` now logs to stdout twice (due to it capturing logs from `Domino` logger). One log comes in cleanly parsed in individual statements and once in raw html form. Not a huge deal (just adds some noise to the logs) but should be an easy fix.

This can be fixed if we either:
- temporarily squelch logging via a context manager
- add the option to disable logging to `runs_start_blocking`

wseaton:
Fixes #63 

Also adds some code to remove `bs4` related warnings by defaulting to the built in parser.

### DIFF
--- a/domino/airflow/_operator.py
+++ b/domino/airflow/_operator.py
@@ -21,7 +21,6 @@ class DominoOperator(BaseOperator):
     discovered via environment variable, as per the domino
     python client.
 
-
     Notes
     ------
     When combining `isDirect=True` with a command, you
@@ -90,8 +89,7 @@ class DominoOperator(BaseOperator):
                     "multipart command string this long if is_direct=True"
                 )
 
-        
-        self.client._logger.setLevel(logging.ERROR)
+        self.client.log.setLevel(logging.ERROR)
         run_response = self.client.runs_start_blocking(
             command=self.command,
             isDirect=self.is_direct,
@@ -102,7 +100,7 @@ class DominoOperator(BaseOperator):
             poll_freq=self.poll_freq,
             max_poll_time=self.max_poll_time
         )
-        self.client._logger.setLevel(logging.INFO)
+        self.client.log.setLevel(logging.INFO)
 
         self.run_id = run_response["runId"]
 
--- a/domino/airflow/_operator.py
+++ b/domino/airflow/_operator.py
@@ -1,4 +1,5 @@
 import time
+import logging
 from typing import List, Optional, Any
 
 from bs4 import BeautifulSoup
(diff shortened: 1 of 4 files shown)


domino/domino.py
```diff
@@ -171,7 +171,7 @@ def runs_start_blocking(self, command, isDirect=False, commitId=None,
                                   .format(run_id))
                     raise Exception(header_msg + stdout_msg)
 
-                logging.info(stdout_msg)
+                self._logger.info(stdout_msg)
```
abhijeet2096, line 174: 👍

domino/airflow/_operator.py
```diff
@@ -89,6 +90,8 @@ def execute(self, context: Any) -> dict:
                     "multipart command string this long if is_direct=True"
                 )
 
+
```
abhijeet2096, line 93: please remove extra blank line

domino/airflow/_operator.py
```diff
@@ -89,6 +90,8 @@ def execute(self, context: Any) -> dict:
                     "multipart command string this long if is_direct=True"
                 )
 
+        
+        self.client._logger.setLevel(logging.ERROR)
```
abhijeet2096, line 94: It would be great if you can add a property in `Domino` class
```
    @property
    def logger(self):
        return self._logger
```
and use it here like
```
self.client.logger.setLevel(logging.ERROR)
```

domino/airflow/_operator.py
```diff
@@ -99,15 +102,18 @@ def execute(self, context: Any) -> dict:
             poll_freq=self.poll_freq,
             max_poll_time=self.max_poll_time
         )
+        self.client._logger.setLevel(logging.INFO)
```
abhijeet2096, line 105: same here

domino/airflow/_operator.py
```diff
@@ -99,15 +102,18 @@ def execute(self, context: Any) -> dict:
...
         log = self.client.get_run_log(self.run_id, self.include_setup_log)
 
         # spool out and replay entire log
         for line in log.splitlines():
-            if line:
+            if line and line != '.':
                 # using bs4 to strip the HTML tags
-                text = BeautifulSoup(line).text
+                # html.parser since it's batteries included
+                text = BeautifulSoup(line, "html.parser").text
```
abhijeet2096, line 116: 👍

domino/airflow/_operator.py
wseaton: Changed to use a property on the main class: `log`

domino/domino.py
```diff
@@ -36,6 +36,14 @@ def __init__(self, project, api_key=None, host=None, domino_token_file=None):
             self._logger.error(error_message)
             raise Exception(error_message)
 
+    @property
+    def log(self):
```
wseaton, line 40: @abhijeet2096 I'm being defensive here because I see a `staticmethod` on the class and you may add more, but realistically this should never not be initialized.

PR #69, opened 2020-08-24
Updating error handling in runs_start_blocking

macrogers:

PR #73, opened 2020-08-28
add hardwaretier validation; adjust run logging

### LINKED ISSUE #67: add hardware tier validation check to domino operator
Now that hardware tiers can be enumerated per project via the API it would be nice if the `DominoOperator` validated that the tier was valid before starting a run and threw a proper exception if invalid (instead of the typical `JSONDecodeError`).

wseaton:
fixes #67, updates logging in accordance with changes from #69.

The way that the `runs_start_blocking` method logs through the exception msg made it hard to catch and handle cleanly, instead I moved around the log processor from the Domino operator and re-factored it to use the new `.log` property which plays nice with airflow. 

Open to suggestions if you guys don't like this behavior, the cost of not fixing it is ugly formatting for logs on airflow job failure, since a pretty printed string is what is passed through the exception (which I then have to de-pretty print), I think ideally we'd get a raw log.

### DIFF
--- a/domino/airflow/_operator.py
+++ b/domino/airflow/_operator.py
@@ -2,11 +2,10 @@ import time
 import logging
 from typing import List, Optional, Any
 
-from bs4 import BeautifulSoup
-
 from airflow.models import BaseOperator
 from airflow.utils.decorators import apply_defaults
 from domino import Domino
+from domino.exceptions import RunFailedException
 
 
 class DominoOperator(BaseOperator):
@@ -38,7 +37,7 @@ class DominoOperator(BaseOperator):
         command: List[str],
         host: Optional[str] = None,
         api_key: Optional[str] = None,
-        domino_token_file: Optional[str] = None, 
+        domino_token_file: Optional[str] = None,
         isDirect: bool = None,
         commitId: Optional[str] = None,
         title: Optional[str] = None,
@@ -54,7 +53,7 @@ class DominoOperator(BaseOperator):
         super(DominoOperator, self).__init__(*args, **kwargs)
 
         self.log.info("Initializing Client...")
-        
+
         self.project = project
         self._api_key = api_key
         self._host = host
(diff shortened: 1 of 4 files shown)


domino/airflow/_operator.py
```diff
@@ -70,14 +69,25 @@ def __init__(
         self.startup_delay = startup_delay
         self.include_setup_log = include_setup_log
 
-        self.client = None
-        self.run_id = None
+        self.client: Optional[Domino] = None
```
abhijeet2096, line 72: 👍

domino/airflow/_operator.py
```diff
@@ -2,11 +2,10 @@
 import logging
```
abhijeet2096, line 2: No longer needed.

domino/airflow/_operator.py
```diff
@@ -2,11 +2,10 @@
...
 from typing import List, Optional, Any
 
-from bs4 import BeautifulSoup
-
 from airflow.models import BaseOperator
 from airflow.utils.decorators import apply_defaults
 from domino import Domino
+from domino.exceptions import RunFailedException
```
abhijeet2096, line 8: unused import

tests/test_operator.py
```diff
@@ -30,3 +30,42 @@ def test_my_operator():
     )
     ti = TaskInstance(task=task, execution_date=datetime.now())
     task.execute(ti.get_template_context())
+
```
abhijeet2096, line 33: Add blank line before `test_operator_fail` and `test_operator_fail_invalid_tier` and at end of file

tests/test_operator.py
```diff
@@ -1,20 +1,20 @@
...
+local mode and run:
+
+`airflow initdb`
+"""
 import os
 from datetime import datetime
 
 import pytest
-
+import logging
```
abhijeet2096, line 11: unused import

abhijeet2096 (approved):
Looks Good, Please fix minor comments

wseaton:
@abhijeet2096 Style changes have been made

PR #76, opened 2020-09-30
[DOM-23013] Use `time_ns` instead of `time` in app_create

abhijeet2096:
This P.R adds following things;
* Exception will be raised for status code 4xx and 5xx.
Screenshot 2020-09-30 at 7 29 36 PM

* Fixes publish app (`created` and `lastUpdated ` values should be given as `long` instead of `float`). Earlier this check was not strict, but as `play-json` was upgraded it has become mandatory.


abhijeet2096:
I have checked this change, and its backwards compatible for Domino version from 4.0 to 4.2

mpkouznetsov (approved):
I don't think time_ns is what is meant to be there but it turns out the server ignores the value. 
We could safely replace it with 0.

It just cares about it being something it can convert to Instant.

PR #77, opened 2020-10-05
[DOM-20304] Fix trailing slash causing init error

abhijeet2096:
This P.R
* Fix trailing slash in case of init error
* Don't log failure response text in case it contains html page
* Add log level environment variable `DOMINO_LOG_LEVEL`


domino/http_request_manager.py
```diff
@@ -31,6 +32,7 @@ def _raise_for_status(self, response):
         try:
             response.raise_for_status()
         except requests.exceptions.HTTPError as e:
-            self._logger.error(e.response.text)
+            if not bool(BeautifulSoup(e.response.text, "html.parser").find()):
```
mpkouznetsov, line 35: Personal preference: I would add a
 ```# Sometimes, the error response is a long HTML page. We don't want to log the whole response html in those cases.```
Also, I would still log it but with `debug`.

mpkouznetsov (approved):
one suggestion.

Issue #70, opened 2020-08-24
runs start via v4 api


wseaton:
Are there plans to add routes and reimplement `runs_start` and `runs_start_blocking` with the v4 API? I'd like to be able to programatically launch runs with a spark cluster attached.

It'd help if there was public documentation on this endpoint so the rest of the params could be implemented, I had to break out my browser dev tools to figure out the following (which should work in 4.2.2):

Example curl:
```bash
curl 'https://domino.corp.company.com/v4/jobs/start' --data-raw
'{"projectId":"5f3ee18bfe89de5d1160890e",
  "commandToRun": "myscript.py",
  "overrideHardwareTierName": "Tiny (M5)",
  "environmentId": "5f3ee2c3fe89de5d11608a1f", 
"onDemandSparkClusterProperties": 
   {"executorCount": 5,
    "executorHardwareTierId": "tinym5",
    "masterHardwareTierId": "tinym5",
    "computeEnvironmentId": "5f3ac75072a3782602bad37e"}
}'
```

abhijeet2096:
Thanks @wseaton , for bringing this up. We are tracking it internally.

ChuckRIHead:
Hey @wseaton,

The docs are auto-generated and we'll be updating them for our next release. 

In the meantime, I want to make sure you're not blocked. Couldn't quite tell from you issue here. 

Do you have a specific question about the API parameters that I can answer in the meantime?

wseaton:
@ChuckRIHead I'm looking for essentially the complete list of v4 params and a short description of what each does. For example, can I provide a raw command to the v4 api? Before I could do this with `isDirect=True`, now that doesn't seem to be an option.

ddl-alexpanin:
@abhijeet2096 ^^^

abhijeet2096:
Hey @wseaton as part of this draft P.R [here](https://github.com/dominodatalab/python-domino/pull/79/files), we are adding two new methods to start job via V4 API. These methods will include option to add `onDemandSparkCluster` to job. Once this gets completed, you can add wrapper method in airflow operator.

abhijeet2096:
Hey @wseaton the P.R (https://github.com/dominodatalab/python-domino/pull/79) got merged today. It should address this issue.

wseaton:
@abhijeet2096 Thanks, I'll work on a new Operator implementation when I get a chance. Appreciate it!

PR #79, opened 2020-10-08
[DOM-23072] Add on demand spark cluster support to domino python package via v4 API to start job

abhijeet2096:
This P.R adds following v4 API endpoints

```
 def job_start(self, command, commit_id=None, hardware_tier_name=None,
                  environment_id=None, on_demand_spark_cluster_properties=None):
 def job_stop(self, job_id, commit_results=True):
 def job_status(self, job_id):
 def job_start_blocking(self, poll_freq=5, max_poll_time=6000, **kwargs):
```

In this P.R i have added a new dependency `[polling2](https://pypi.org/project/polling2/)`, for blocking start job method.

### Linked issues
#70 

### References
https://dominodatalab.atlassian.net/browse/DOM-23072

### DIFF
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -366,10 +366,22 @@ class Domino:
         ).json()
 
     def job_start_blocking(self, poll_freq=5, max_poll_time=6000, **kwargs):
+        """
+        Starts a job in a blocking loop, periodically polling for status of job
+        is complete. Will ignore intermediate request exception.
+        :param poll_freq: int (Optional) polling frequency interval in seconds
+        :param max_poll_time: int (Optional) max poll time in seconds
+        :param kwargs: Additional arguments to be passed to job_start
+        :return:
+        """
+        def get_job_status(job_identifier):
+            status = self.job_status(job_identifier)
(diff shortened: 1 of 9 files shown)


domino/domino.py
```diff
@@ -243,6 +234,164 @@ def runs_stdout(self, runId):
         # pprint.pformat outputs a string that is ready to be printed
         return pprint.pformat(self._get(url)['stdout'])
 
+    def job_start(self, command, commit_id=None, hardware_tier_name=None,
```
mpkouznetsov, line 237: Is it possible to have a return type hint here, even if it is a dict? (Nice to have)

mpkouznetsov (approved):
Can you describe how this has been tested? I understand that testing requires a server instance so integration tests would not be easy to implement.

abhijeet2096:
Hey @mpkouznetsov currently we don't have any unit tests for `python-domino`. I did manually tested these methods against 4.3 and 4.2 releases. I think we have plans to incorporate system-tests for `python-domino`. For older methods we have manual tests in test rail, I will add more tests for these new methods in test rail.

PR #80, opened 2020-10-22
DominoSparkOperator

wseaton:
Add new `DominoSparkOperator` that supports the v4 job start function signature along with associated unit tests.

### DIFF
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -377,6 +377,8 @@ class Domino:
         def get_job_status(job_identifier):
             status = self.job_status(job_identifier)
             self.log.info(f"Polling Job: {job_identifier} status is completed: {status['statuses']['isCompleted']}")
+            if status['statuses']['executionStatus'] == "Failed":
+                raise RunFailedException(f"Run Status Failed for run: {job_identifier}")
             return status
         job = self.job_start(**kwargs)
         job_id = job['id']
--- a/tests/test_spark_operator.py
+++ b/tests/test_spark_operator.py
@@ -52,3 +52,22 @@ def test_spark_operator_with_cluster():
     )
     ti = TaskInstance(task=task, execution_date=datetime.now())
     task.execute(ti.get_template_context())
+
+    
+def test_spark_operator_no_cluster_failed():
+    airflow = pytest.importorskip("airflow")
+
+    from airflow import DAG
+    from airflow.models import TaskInstance
+
+    dag = DAG(dag_id="foo", start_date=datetime.now())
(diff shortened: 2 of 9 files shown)


wseaton:
@abhijeet2096  For my personal development environment I've been using poetry to manage package dependencies (also to pull in test dependencies like pytest), I can also push those files if you all are interested.

domino/domino.py
```diff
@@ -377,6 +377,8 @@ def job_start_blocking(self, poll_freq: int = 5, max_poll_time: int = 6000, **kw
         def get_job_status(job_identifier):
             status = self.job_status(job_identifier)
             self.log.info(f"Polling Job: {job_identifier} status is completed: {status['statuses']['isCompleted']}")
+            if status['statuses']['executionStatus'] == "Failed":
```
wseaton, line 380: @abhijeet2096 throwing an exception here if the status is failed so it can be handled downstream by other functions, like the airflow operator. I noticed failures were happening silenty since we just check for job completion and
`job_complete != job_success`

tests/test_spark_operator.py
```diff
@@ -0,0 +1,73 @@
...
+
+    from airflow import DAG
+    from airflow.models import TaskInstance
+
+    dag = DAG(dag_id="foo", start_date=datetime.now())
+    task = DominoSparkOperator(
+        dag=dag,
+        task_id="foo",
+        project=TEST_PROJECT,
+        command="test_spark.py",
```
abhijeet2096, line 30: Hey @wseaton can we have python/bash commands instead of name of the test files `test_spark.py`, `test_spark_fail.sh`?

abhijeet2096 (commented):
This looks good. Can you also update the associated readme for airflow section.

tests/test_spark_operator.py
wseaton: Not here, since I can't provide a direct command via the v4 api, before I could do this via `isDirect`. I'd also like to not have to use wrapper scripts :slightly_smiling_face:

wseaton:
@abhijeet2096 readme has been updated

tests/test_spark_operator.py
abhijeet2096: Hey @wseaton can you add these sample test files (test_spark.py, test_spark_fail.sh) here, in this folder?

wseaton: @abhijeet2096 this has been done

abhijeet2096 (approved):
🚀

PR #81, opened 2020-11-04
Check if the App is running before stopping

mpkouznetsov:
The domino server responds with 500 when we attempt to stop an App that is already stopped.

Note: this change does not completely eliminate the possibility of a race condition when another client stops an App between the time we check its status and issue a stop request.


abhijeet2096 (approved):
Looks Good!

ddl-alexpanin:
@mpkouznetsov can you please add a note to `CHANGELOG.md`?

PR #83, opened 2021-01-05
[DOM-20439] Remove name from Model Version Publish

mohithg:


abhijeet2096 (approved):
Looks good. Can you also update the [Changelog](https://github.com/dominodatalab/python-domino/blob/master/CHANGELOG.md).

PR #86, opened 2021-03-22
Add compute cluster properties to job launching

niole:
This PR adds compute cluster support to the start job endpoint. This looks like it is the only endpoint that can launch cluster related executions, so I only added it there.

There doesn't seem to be a standardized way for doing input validation, yet there is input validation on the start job endpoint, so I added basic input validation for the compute cluster definition. I did not do validation on whether the HWTs or env ids exist or not, because domino will reject with informative messaging if that is the case.

https://dominodatalab.atlassian.net/browse/DOM-27821

### DIFF
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -241,11 +241,11 @@ class Domino:
     def job_start(
             self,
             command: str,
-            commit_id: str = None,
-            hardware_tier_name: str = None,
-            environment_id: str = None,
-            on_demand_spark_cluster_properties: dict = None,
-            compute_cluster_properties: dict = None) -> dict:
+            commit_id: Optional[str] = None,
+            hardware_tier_name: Optional[str] = None,
+            environment_id: Optional[str] = None,
+            on_demand_spark_cluster_properties: Optional[dict] = None,
+            compute_cluster_properties: Optional[dict] = None) -> dict:
         """
         Starts a Domino Job via V4 API
         :param command:                             string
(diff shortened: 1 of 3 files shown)


domino/domino.py
```diff
@@ -255,8 +261,9 @@ def job_start(self, command: str, commit_id: str = None, hardware_tier_name: str
                                                     The environment id to launch job with. If not provided
                                                     it will use the default environment for the project
         :param on_demand_spark_cluster_properties:  dict (Optional)
-                                                    On demand spark cluster properties. Following properties
-                                                    can be provided in spark cluster
+                                                    On demand spark cluster properties. Deprecated by
+                                                    compute_cluster_properties. Please use compute_cluster_properties
```
mpkouznetsov, line 265: The "by" clause seems redundant and grammatically questionable.
```suggestion
                                                    On demand spark cluster properties. Deprecated.
                                                    Please use compute_cluster_properties
```

domino/domino.py
```diff
@@ -238,8 +238,14 @@ def runs_stdout(self, runId):
...
-    def job_start(self, command: str, commit_id: str = None, hardware_tier_name: str = None,
-                  environment_id: str = None, on_demand_spark_cluster_properties: dict = None) -> dict:
+    def job_start(
+            self,
+            command: str,
+            commit_id: str = None,
+            hardware_tier_name: str = None,
+            environment_id: str = None,
+            on_demand_spark_cluster_properties: dict = None,
+            compute_cluster_properties: dict = None) -> dict:
```
mpkouznetsov, line 248: Not super-important but more "thorough" (see https://docs.python.org/3/library/typing.html#typing.Optional)
 ```suggestion
             commit_id: Optional[str] = None,
             hardware_tier_name: Optional[str] = None,
             environment_id: Optional[str] = None,
             on_demand_spark_cluster_properties: Optional[dict] = None,
             compute_cluster_properties: Optional[dict] = None) -> dict:
```

mpkouznetsov (approved):
Looks good; two minor comments.

niole:
Looks like since this PR was created, we merged leo's revision spec work. So I will be updating the start job api to reflect that

domino/domino.py
niole: yes, this is a good idea

domino/domino.py
niole: ok, updated

domino/domino.py
niole: done

chittshota (changes requested):
@niole Changes look fine other than the master storage option. Can you remove it completely here. masterStorage is becoming obsolete. We can remove it here and later on from the backend as well. For now this API doesn't need to send any value for masterStorage or None/Non-validated.

PR #91, opened 2021-05-20
fix add collaborators_add to work with v4 api

ddl-abayly:
Problem:
Because of recent changes to the server, collaborators_add() no longer works. 

Solution:
re-work collaborators_add() to use the v4 API instead, thus achieving compatibility with both versions 4.5 and 4.4.
collaborators_remove has been added as well to facilitate testing.

DOM-29690


ddl-alexpanin (approved):
LGTM

domino/domino.py
```diff
@@ -525,15 +527,38 @@ def collaborators_get(self):
         url = self._routes.collaborators_get()
         return self._get(url)
 
+    def get_userId(self, usernameOrEmail):
```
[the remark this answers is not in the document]
ddl-abayly: fixed.

domino/domino.py
```diff
@@ -525,15 +527,38 @@ def collaborators_get(self):
         url = self._routes.collaborators_get()
         return self._get(url)
 
+    def get_userId(self, usernameOrEmail): 
+        url = self._routes.users_get()
+        response = self.request_manager.get(url)
+        users = json.loads(response.text)
```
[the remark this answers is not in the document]
ddl-abayly: fixed.

mpkouznetsov (approved):
a couple of suggestions.

dknupp:
Sorry -- I have some additional comments here, but I'm juggling a few things at the moment. Mainly, I'm concerned about what happens in the case that no user is found -- what it is desired behavior. As written, this code is just going to throw some user-hostile exceptions.

There are still a couple of python considerations as well.

ddl-abayly:
added friendly message if user is not found: 

  File "/Users/andrewbayly/dev/python-domino/tests/test_collaborators.py", line 11, in <module>
    test_collaborators()    
  File "/Users/andrewbayly/dev/python-domino/tests/test_collaborators.py", line 6, in test_collaborators
    print( domino.collaborators_add('user22') )
  File "/Users/andrewbayly/dev/python-domino/domino/domino.py", line 544, in collaborators_add
    user_id = self.get_user_id(username_or_email)
  File "/Users/andrewbayly/dev/python-domino/domino/domino.py", line 536, in get_user_id
    raise Exception(f"Could not find user: {username_or_email}")
Exception: Could not find user: user22

domino/domino.py
```diff
@@ -525,15 +525,41 @@ def collaborators_get(self):
...
+        response = self.request_manager.get(url)
+        users = response.json()
+        user_from_user_name = [x for x in users if ('userName' in x) and (x['userName'] == username_or_email)]
+        user_from_email = [x for x in users if ('email' in x) and (x['email'] == username_or_email)]
+        user_list = user_from_user_name + user_from_email
+        if len(user_list) == 0:
+            raise Exception(f"Could not find user: {username_or_email}")
+        user = user_list[0]
+        userId = user['id']
+        return userId
```
dknupp, line 539: Since this is a public-facing method (i.e., not just test code), I think we need to be sure what the correct response should be in the case that the user isn't found. Throwing a bare `Exception` is probably not the best choice. My guess is that returning `None` is more appropriate. @mpkouznetsov or @ddl-mmahmoud?

To save time on the PR, I would suggest replacing the method with this.
```python
def get_user_id(self, username_or_email):
    """
    Return the user ID of the the user account matching the given search criteria.

    Args:
        username_or_email (str): the username or email account of a user

    Return:
        user_id (str): the matching user ID, or None if not found
    """
    url = self._routes.users_get()
    response = self.request_manager.get(url)
    users = response.json()

    user_id = None
    for user in users:
        if username_or_email in (user['userName'], user['email']):
            user_id = user['id']
            break
    return user_id
```
I grabbed two default users from a recent deployment...
```python
users = [
  {
    "firstName": "integration-test",
    "lastName": "integration-test",
    "fullName": "integration-test integration-test",
    "userName": "integration-test",
    "email": "test-notifs+integration-test@dominodatalab.com",
    "avatarUrl": "",
    "id": "60a301d1049cc4002e81c7c8"
  },
  {
    "firstName": "Jira",
    "lastName": "User",
    "fullName": "Jira User",
    "userName": "jira_user_anonymous",
    "email": "jira@dominodatalab.com",
    "avatarUrl": "",
    "id": "60a301e14d3bd8efde3eb94b"
  }
]
```
...then tested the following conditions, and they all seem to pass.
```python
# Test search by valid email
u0 = get_user_id(None, 'test-notifs+integration-test@dominodatalab.com')
assert u0 == '60a301d1049cc4002e81c7c8'

# Test search by valid username
u1 = get_user_id(None, 'jira_user_anonymous')
assert u1 == '60a301e14d3bd8efde3eb94b'

# Test non-existent user
u2 = get_user_id(None, 'foo')
assert u2 == None

# Test username substring does not return false positive
u3 = get_user_id(None, 'integra')
assert u3 == None

# Test email substring does not return false positive
u4 = get_user_id(None, '@dominodatalab.com')
assert u4 == None
```

domino/routes.py
```diff
@@ -151,3 +151,7 @@ def hardware_tiers_list(self, project_id):
     def find_project_by_owner_name_and_project_name_url(self):
         return f'{self.host}/v4/gateway/projects/findProjectByOwnerAndName' \
                f'?ownerName={self._owner_username}&projectName={self._project_name}'
+
+    # User URLs
+    def users_get(self):
+        return self.host + f'/v4/users'
```
dknupp, line 157: This is beyond the scope of this PR obviously, but it makes me sad that we don't have an API to return a single user from the DB, but instead have to retrieve all users on the system, then search though them in application code.

dknupp (changes requested):

domino/domino.py
```diff
@@ -525,15 +525,41 @@ def collaborators_get(self):
...
         return self._get(url)
 
-    def collaborators_add(self, usernameOrEmail, message=""):
+    def get_user_id(self, username_or_email): 
+        url = self._routes.users_get()
+        response = self.request_manager.get(url)
+        users = response.json()
+        user_from_user_name = [x for x in users if ('userName' in x) and (x['userName'] == username_or_email)]
+        user_from_email = [x for x in users if ('email' in x) and (x['email'] == username_or_email)]
+        user_list = user_from_user_name + user_from_email
```
dknupp, line 534: Apologies for not explaining my thoughts in the longer comment earlier @ddl-abayly -- I was trying to finish up two other tasks, and wanted to post something quickly in case you were waiting to merge something by the end of the day. With a bit more time, I can expand on it a bit. (Also, with hindsight, I fixed a glaring bug in my earlier suggestion.)

I think the three lines here are doing more work than you necessary. 

* Each list comprehension iterates through the `users` list independently, while the `for` loop iterates through it only once.
* You don't need to check "if key in user" if you use `user.get(key)`. (This was my bug. I didn't realize that some users don't have email accounts -- like the default Jira user. Accessing the key like this won't throw a `KeyError`.)
* The original approach creates two (albeit short, or even empty) lists, and then a third by concatenating those, just to get at the first element -- which all seemed busy, and a bit hard to read.

Of course, my suggested edit assumes that users have unique usernames and email address (which seems like a reasonable assumption).

domino/domino.py
```diff
@@ -525,15 +525,41 @@ def collaborators_get(self):
...
+        if len(user_list) == 0:
+            raise Exception(f"Could not find user: {username_or_email}")
```
dknupp, line 536: With regard to the expected behavior when a user is not present, I don't have authoritative experience on the subject, but I do know that other similar API's that provide a similar lookup (e.g., Keycloak's API also lets you look up a user ID by user name, or even just any basic DB lookup) doesn't throw an exception if a record matching the given search term is not found. Typically, returning None or an empty set is sufficient. Throwing an Exception seems unexpected to me.

domino/domino.py
```diff
@@ -525,15 +525,41 @@ def collaborators_get(self):
...
+        return userId
+
+    def collaborators_add(self, username_or_email, message=""):
         self.requires_at_least("1.53.0.0")
-        url = self._routes.collaborators_add()
-        data = {
-            'collaboratorUsernameOrEmail': usernameOrEmail,
-            'welcomeMessage': message
+ 
+        user_id = self.get_user_id(username_or_email)
```
dknupp, line 544: On the other hand, I think it would be reasonable to raise an exception in the calling method here if trying to add a non-existent user as a collaborator -- but rather than a bare base Exception, we should define a new exception class in [domino/exceptions.py](https://github.com/dominodatalab/python-domino/blob/master/domino/exceptions.py), something like:
```python
class UserNotFoundException(DominoException):
    """User not found Exception"""
    pass
````
...and then raise it here with the friendly message.
```suggestion
        user_id = self.get_user_id(username_or_email)
        if user_id is None:
            raise UserNotFoundException(f"Could not add collaborator matching {username_or_email}")
```

domino/domino.py
```diff
@@ -525,15 +525,41 @@ def collaborators_get(self):
...
+
+        response = self.request_manager.post(url, json=request)
+        return response
+
+    def collaborators_remove(self, username_or_email):
+        self.requires_at_least("1.53.0.0")
+
+        userId = self.get_user_id(username_or_email)
+    
+        url = self._routes.collaborators_remove(self._project_id, userId)
```
dknupp, line 560: The same thing here.
```suggestion
        user_id = self.get_user_id(username_or_email)
        if user_id is None:
            raise UserNotFoundException(f"Could not remove collaborator matching {username_or_email}")
    
        url = self._routes.collaborators_remove(self._project_id, user_id)
```

domino/domino.py
```diff
@@ -525,15 +525,38 @@ def collaborators_get(self):
         url = self._routes.collaborators_get()
         return self._get(url)
 
-    def collaborators_add(self, usernameOrEmail, message=""):
+    def get_user_id(self, username_or_email): 
+        url = self._routes.users_get()
+        response = self.request_manager.get(url)
+        users = response.json()
+        user_from_user_name = [x for x in users if ( ('userName' in x ) and ( x['userName'] == username_or_email ) ) ]
+        user_from_email = [x for x in users if ( ( 'email' in x ) and ( x['email'] == username_or_email ) ) ]
```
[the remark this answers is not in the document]
ddl-abayly: fixed.

domino/domino.py
ddl-abayly: fixed as suggested.

domino/domino.py
ddl-abayly: fixed.

domino/domino.py
ddl-abayly: fixed.

dknupp:
Thanks for making all of those last minute changes @ddl-abayly

PR #101, opened 2021-07-29
Add, validate, and document autoscaling configuration for jobs_start

niole:
This PR adds the ability to start jobs with compute clusters that can autoscale.

https://dominodatalab.atlassian.net/browse/DOM-31483

### DIFF
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -331,10 +331,6 @@ class Domino:
             default_spark_setting_url = self._routes.default_spark_setting(self._project_id)
             return self.request_manager.get(default_spark_setting_url).json()
 
-        def validate_autoscaling_supported():
-            if not is_comute_cluster_autoscaling_supported(self._version):
-                raise Exception(f"Domino {self._version} does not support compute cluster autoscaling.")
-
         def validate_is_on_demand_spark_supported():
             if not is_on_demand_spark_cluster_supported:
                 raise OnDemandSparkClusterNotSupportedException(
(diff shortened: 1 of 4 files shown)


domino/domino.py
```diff
@@ -327,6 +331,10 @@ def get_default_spark_settings():
             default_spark_setting_url = self._routes.default_spark_setting(self._project_id)
             return self.request_manager.get(default_spark_setting_url).json()
 
+        def validate_autoscaling_supported():
+            if not is_comute_cluster_autoscaling_supported(self._version):
+                raise Exception(f"Domino {self._version} does not support compute cluster autoscaling.")
```
chittshota, line 336: am wondering if there is a better approach here. If version is not >= 5.0.0 instead of throwing the exception and disallowing user (rather forcing user to remove maxcount), we can print warning that this value will be ignored and cluster wont auto-scale. However a static cluster with workerCount will be created.

niole: I'm not sure that it is better to warn the user and then proceed with only workerCount. If we only warn and then proceed, we are assuming that the user's `workerCount` selection is good enough to use for the workload that they want to do. I guess that I am assuming that users put a lot of thought into their worker selections, which might not be the case.

Wdyt: on average, would it be more inconvenient to the user if we proceeded with launching an execution that might end up not having enough resources?

chittshota: Don't have a strong opinion either way. Your point is also quite pertinent. Let's just stick to the original model of simply throwing error and disallowing the flow if maxworker is populated.

dknupp: For what it's worth, I agree that failing early with a clear message, rather than potentially failing later for a less than obvious reason, is a much better approach.

On a related note, I see that there is a well-established pattern here of raising the base `Exception` class, but I believe this is generally frowned upon. E.g., static code analysis tools will commonly [flag this](https://rules.sonarsource.com/python/RSPEC-112). Since there's an `exceptions.py`  file already, how would you feel about creating a specific exception type? I do see an `OnDemandSparkClusterNotSupportedException` as the one counter example in this method.

I'm not even sure that we need something that specific for auto-scaling -- but being a little more descriptive might be helpful. `FeatureNotSupported` or `InvalidJobParameter` or something. Even a base `DominoException` would be better.

niole: sure, I'll see what I can come up with...

niole: ok @dknupp ptal

dknupp (approved):
Thanks @niole for doing that -- it looks good to me.

chittshota:
Looks good. Am unable to approve this.

PR #103, opened 2021-08-19
Release 1.0.5

ddl-alexpanin:
Release 1.0.5.


domino/_version.py
```diff
@@ -1 +1 @@
-__version__ = '1.0.4'
+__version__ = '1.0.5'
```
dknupp, line 1: A discussion with Med in a Slack thread came up re: changing the version scheme of this package to match the corresponding Domino version. A user would know immediately, for example, that to access a API features associated Domino 4.5.0, they would need to pip install `python-domino==4.5.0`.

@mpkouznetsov @dmitriyvolk -- any thoughts?

dknupp:
> Also will be publishing releases to PyPI.

One general thing to note @ddl-alexpanin about PyPI. If you release a package versioned as 1.2.3, that number can never be re-used. In other words, if you notice an error after the fact, correct it, and then try to re-upload version 1.2.3, you will be prevented from doing so.

Any change, even if it's just a typo fix in the README, will require a new version (e.g., 1.2.4).

This is a security feature so that no package can ever be surreptitiously replaced with code that mimics it, but also perhaps does something harmful.

ddl-alexpanin:
> One general thing to note @ddl-alexpanin about PyPI. If you release a package versioned as 1.2.3, that number can never be re-used. 

Yes, this is why https://test.pypi.org/project/python-domino/ is already 1.0.13 :)

dknupp:
> Yes, this is why https://test.pypi.org/project/python-domino/ is already 1.0.13 :)

👍

domino/_version.py
mpkouznetsov: No strong opinion. python-domino supports multiple versions of domino so aligning the versioning schemes will sort of hint that we support _only_ the latest version. Still, it probably makes more sense to align.

ddl-alexpanin: Discussed with Pablo and we decided not to change versioning scheme at this moment.

mpkouznetsov (approved):
LGTM

PR #105, opened 2021-09-09
Automatically re-read auth token from file to avoid reusing short-lived access tokens

ddl-alexpanin:
Domino's python wrapper has `BearerAuth` class which reads the token from file during instantiation, and then keeps reusing it for all the subsequent calls. This doesn't work well with short-lived access tokens provided by JWT credentials propagation, when the token expires. Since Domino 4.5.0, access tokens are short-lived (5 minutes) by default.

Modifying `BearerAuth` to re-read token from file any time API request is made.

Note as there were issues with pod readiness in Domino 4.5 and below, this fix won't likely prevent authentication issues in 4.5.0 due to short-lived token expiration.


dmitriyvolk (approved):
Yep, this looks exactly right, although I would feel better if there was a test for that, but every time I have to use a mocker in python, I cry :(

dknupp:
I re-ran the existing unit tests, and they all pass.

If there's no objection, I'd like a day or two to think about other tests, possible corner cases, or unit tests that could be added.
```
================================= test session starts ================================= 
platform linux -- Python 3.6.9, pytest-6.2.5, py-1.10.0, pluggy-1.0.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /mnt/python-domino, configfile: pytest.ini
plugins: requests-mock-1.9.3
collected 11 items                                                                                                                                             

test_basic_auth.py::test_no_auth_type_error PASSED
test_basic_auth.py::test_invalid_token_file_error PASSED
test_basic_auth.py::test_malformed_project_input_error PASSED
test_basic_auth.py::test_object_creation_with_auth_token PASSED
test_basic_auth.py::test_object_creation_with_api_key PASSED
test_basic_auth.py::test_object_creation_with_token_file PASSED
test_basic_auth.py::test_re_authentication_with_a_new_type PASSED
test_basic_auth.py::test_auth_token_precedence PASSED
test_basic_auth.py::test_token_file_precedence PASSED
test_basic_auth.py::test_auth_against_real_deployment_with_api_key PASSED
test_basic_auth.py::test_auth_against_real_deployment_with_token_file PASSED
```

ddl-alexpanin:
> If there's no objection, I'd like a day or two to think about other tests, possible corner cases, or unit tests that could be added.

Thanks! Of course, this could wait until next week.

dknupp (approved):
Thanks for the fix, @ddl-alexpanin. This looks good to me.

PR #106, opened 2021-10-05
Better error if auth token needs refresh when running jobs

ddl-alexpanin:
Added better token expiration error handling for `job_start` and `runs_start` methods (the latter is not really necessary right now as it uses v1 API).

### DIFF
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -96,9 +96,11 @@ class Domino:
             "tier": tier,
             "publishApiEndpoint": publishApiEndpoint
         }
-
-        response = self.request_manager.post(url, json=request)
-        return response.json()
+        try:
+            response = self.request_manager.post(url, json=request)
+            return response.json()
+        except ReloginRequiredException:
+            self._logger.info(f" You need to log in to the Domino UI to start the run. Please do it at {self._routes.host}")
 
     def runs_start_blocking(self, command, isDirect=False, commitId=None,
                             title=None, tier=None, publishApiEndpoint=None,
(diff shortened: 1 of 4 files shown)


domino/http_request_manager.py
```diff
@@ -33,6 +36,9 @@ def _raise_for_status(self, response):
         try:
             response.raise_for_status()
         except requests.exceptions.HTTPError as e:
+            if e.response.status_code == HTTPStatus.CONFLICT:
+                parse_result = urlparse(e.response.url)
+                self._logger.error(f" An error has occurred. Try to relogin at {parse_result.scheme}://{parse_result.netloc}")
```
dmitriyvolk, line 41: I don't think it's an error condition. I would treat is as a normal outcome and say something like "In order to start your _execution_ you need to login to the Domino UI, please do so at <url>"

But in this case you'd have to throw a _very_ specific exception here, and catch it in  starting job and staring workspace.

domino/domino.py
```diff
@@ -96,9 +96,11 @@ def runs_start(self, command, isDirect=False, commitId=None, title=None,
...
             "publishApiEndpoint": publishApiEndpoint
         }
-
-        response = self.request_manager.post(url, json=request)
-        return response.json()
+        try:
+            response = self.request_manager.post(url, json=request)
+            return response.json()
+        except ReloginRequiredException:
+            self._logger.info(f" You need to log in to the Domino UI to start the run. Please do it at {self._routes.host}")
```
dmitriyvolk, line 103: I would slightly modify the url to be something like `{self._routes.host}/relogin?redirectPath=/

PR #108, opened 2021-10-21
Enable edv support to start job

ronit-basu:
We currently don't have validation for the external volume mount ids nor do we have an easy way of validating them. Need to sync with someone familiar with this repo to learn best practices here

This PR is from a fork since i don't have write access. Once i get write access i will close this in favor of a new one that follows the normal approach

### DIFF
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -252,7 +252,9 @@ class Domino:
             hardware_tier_name: Optional[str] = None,
             environment_id: Optional[str] = None,
             on_demand_spark_cluster_properties: Optional[dict] = None,
-            compute_cluster_properties: Optional[dict] = None) -> dict:
+            compute_cluster_properties: Optional[dict] = None,
+            external_volume_mounts: Optional[List[str]] = None,
+    ) -> dict:
         """
         Starts a Domino Job via V4 API
         :param command:                             string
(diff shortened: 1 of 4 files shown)


domino/domino.py
```diff
@@ -374,6 +379,12 @@ def throw_if_information_invalid(key: str, info: dict) -> bool:
             if "maxWorkerCount" in compute_cluster_properties and not is_comute_cluster_autoscaling_supported(self._version):
                 raise UnsupportedFieldException(f"'maxWorkerCount' is not supported in Domino {self._version}.")
 
+        def validate_is_external_volume_mounts_supported():
+            if not is_external_volume_mounts_supported:
```
ddl-alexpanin, line 383: Where this variable is defined?

ddl-alexpanin (commented):
Also need to:

1. Update `README.md` with the information about new parameter in `job_start`
2. Update `CHANGELOG.md`

domino/domino.py
ronit-basu: it is defined in domino/helpers.py

ddl-alexpanin: I just do not understand:

1. `is_external_volume_mounts_supported(version: str) -> bool` in `helpers.py` is a function.
2. What does `not is_external_volume_mounts_supported` mean?

PR #116, opened 2021-12-14
Added support for listOnlyUsers to users_get

ddl-awroblicky:


ddl-alexpanin (commented):
Guess these are routes, do we want to have public facing methods for getting users and orgs?

ddl-awroblicky:
@ddl-alexpanin I'm not sure I can close if this is not something we want to have public facing

PR #118, opened 2022-01-24
Update domino to accept version as a parameter

hao-fast:
the version API is in the direct domain host without `/v1` or `/v4` path, this will block the API access if configured with SSO.
This fix will accept version as a parameter when using the API

### DIFF
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -20,7 +20,7 @@ from bs4 import BeautifulSoup
 
 
 class Domino:
-    def __init__(self, project, api_key=None, host=None, domino_token_file=None, auth_token=None):
+    def __init__(self, project, version=None, api_key=None, host=None, domino_token_file=None, auth_token=None):
 
         self._configure_logging()
 
@@ -43,7 +43,10 @@ class Domino:
         self.authenticate(api_key, auth_token, domino_token_file)
 
         # Get version
-        self._version = self.deployment_version().get("version")
+        if version:
+            self._version = version
+        else:
+            self._version = self.deployment_version().get("version")
         self._logger.info(f"Domino deployment {host} is running version {self._version}")
 
         # Check version compatibility


ddl-alexpanin:
@hao-fast can you please describe your use case in details, what exactly are you trying to achieve and what does not work.

We can't allow passing version this way as the version passed should be the same as the version of used Domino instance, if they won't match the correctness of internal behavior of the library is not guaranteed.

hao-fast:
> @hao-fast can you please describe your use case in details, what exactly are you trying to achieve and what does not work.
> 
> 
> 
> We can't allow passing version this way as the version passed should be the same as the version of used Domino instance, if they won't match the correctness of internal behavior of the library is not guaranteed.

@ddl-alexpanin So the reason is the version api do not starts with /v1 or /v4. Our SSO will block it direct response without those prefix in path.
If you could move it to /v1/version that will be awesome

dmitriyvolk:
Hi @hao-fast ! 
Could you please provide some more details about your usecase?
I don't quite understand the connection between SSO and path to the `version` API.

hao-fast:
> Hi @hao-fast ! Could you please provide some more details about your usecase? I don't quite understand the connection between SSO and path to the `version` API.

So our cloudflare require SSO for any domino direct path which means domino_host/version requires SSO. you can think of any domino_host/* needs you to sign in through okta.
and our Cloudflare doesn't require API endpoint(domino_host/v1 or domino_host/v4) to do SSO.

ddl-alexpanin:
> > Hi @hao-fast ! Could you please provide some more details about your usecase? I don't quite understand the connection between SSO and path to the `version` API.
> 
> So our cloudflare require SSO for any domino direct path which means domino_host/version requires SSO. you can think of any domino_host/* needs you to sign in through okta. and our Cloudflare doesn't require API endpoint(domino_host/v1 or domino_host/v4) to do SSO.

@hao-fast the `/version` endpoint actually does not require any authentication at all, check our "trial public cloud" instance: https://try.dominodatalab.com/version . Maybe your Cloudflare admins can add an exception for this endpoint based on this information?

PR #121, opened 2022-04-05
Fixing precedence of authentication methods

ddl-alexpanin:
When instantiating `Domino` object, the authentication preference should always be given to the authentication token. If it's not passed, the path to domino token file takes precedence, otherwise the API key is used. If none of these three parameters are passed, then preference will be given to the domino token file from the corresponding environment variable, then to the API key from the corresponding environment variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ You can set up the connection by creating new instance of `Domino`.
 
 <hr>
 
-### *class* Domino(project, api_key=None, host=None, domino_token_file=None)
+### *class* Domino(project, api_key=None, host=None, domino_token_file=None, auth_token=None)
 
 The parameters are:
 
@@ -51,9 +51,13 @@ The parameters are:
   environment variable.
 * *domino_token_file:* (Optional) Path to domino token file containing auth token. If not provided the library will expect to find one
 in the DOMINO_TOKEN_FILE environment variable.
+* *auth_token:* (Optional) Authentication token
 
 Note:
-1. In case both api_key and domino_token_file are available, then preference will be given to domino_token_file.
+1. The preference is always given to authentication token. If it's not passed, the path to domino token file
+takes precedence, if it is explicitly passed as a parameter, otherwise the API key is used. If none of these three
+parameters is passed,  then preference will be given to domino_token_file from the environment variable, then
+to the API key from the environment variable
 2. By default the log level is set to `INFO`, to set log level to `DEBUG`, set `DOMINO_LOG_LEVEL` environment variable to `DEBUG`
 <hr>
 
--- a/README.md
+++ b/README.md
@@ -54,10 +54,9 @@ in the DOMINO_TOKEN_FILE environment variable.
 * *auth_token:* (Optional) Authentication token
 
 Note:
-1. The preference is always given to authentication token. If it's not passed, the path to domino token file
-takes precedence, if it is explicitly passed as a parameter, otherwise the API key is used. If none of these three
-parameters is passed,  then preference will be given to domino_token_file from the environment variable, then
-to the API key from the environment variable
+1. The authentication preference should always be given to the authentication token. If it's not passed, the path to domino token file
+takes precedence, otherwise the API key is used. If none of these three parameters are passed, then preference will be given to the
+domino token file from the corresponding environment variable, then to the API key from the corresponding environment variable.
 2. By default the log level is set to `INFO`, to set log level to `DEBUG`, set `DOMINO_LOG_LEVEL` environment variable to `DEBUG`
 <hr>
 
--- a/domino/authentication.py
+++ b/domino/authentication.py
@@ -1,5 +1,9 @@
 from requests.auth import AuthBase, HTTPBasicAuth
 
+from .constants import *
+
+import os
+
 
 class BearerAuth(AuthBase):
     """
@@ -26,7 +30,7 @@ class BearerAuth(AuthBase):
         return r
 
 
-def get_auth_by_type(api_key=None, auth_token=None, domino_token_file=None, api_key_from_env=None, domino_token_file_from_env=None):
+def get_auth_by_type(api_key=None, auth_token=None, domino_token_file=None):
     """
     Return appropriate authentication object for requests.
 
@@ -40,9 +44,6 @@ def get_auth_by_type(api_key=None, auth_token=None, domino_token_file=None, api_
         5. api_key_from_env
     """
 
-    assert any([api_key, auth_token, domino_token_file, api_key_from_env, domino_token_file_from_env]), \
-        "Unable to authenticate: no authentication method provided"
-
     if auth_token is not None:
         return BearerAuth(auth_token=auth_token)
     elif domino_token_file is not None:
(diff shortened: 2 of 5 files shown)


domino/domino.py
```diff
@@ -36,11 +36,21 @@ def __init__(self, project, api_key=None, host=None, domino_token_file=None, aut
...
+        actual_domino_token_file = None
+
+        if not auth_token:
+            if domino_token_file:
+                actual_domino_token_file = domino_token_file
+            elif api_key:
+                actual_api_key = api_key
+            else:
+                actual_api_key = os.getenv(DOMINO_USER_API_KEY_KEY_NAME)
+                actual_domino_token_file = os.getenv(DOMINO_TOKEN_FILE_KEY_NAME)
```
dmitriyvolk, line 50: I would like to make sure that when these things are used, the right precedence is used as well (token > api key).

tests/test_basic_auth.py
```diff
@@ -124,6 +124,21 @@ def test_auth_token_precedence():
         "With multiple authentication options, auth_token string should take highest precendence"
 
 
+@pytest.mark.usefixtures("mock_domino_version_response")
+@pytest.mark.skipif(not os.getenv(DOMINO_TOKEN_FILE_KEY_NAME), reason="No token file in environment")  # noqa:E501
+def test_auth_with_only_api_key_passed_as_parameter():
```
dmitriyvolk, line 129: Maybe we can have ONE test that literally follows the table from the ticket?

dmitriyvolk (commented):
The logic looks right, but ideally I would really like all this logic to be a part of the same function, rather than scattered between auth.py and domino.py. But I believe what you have here works

README.md
```diff
@@ -51,9 +51,12 @@ The parameters are:
   environment variable.
 * *domino_token_file:* (Optional) Path to domino token file containing auth token. If not provided the library will expect to find one
 in the DOMINO_TOKEN_FILE environment variable.
+* *auth_token:* (Optional) Authentication token
 
 Note:
-1. In case both api_key and domino_token_file are available, then preference will be given to domino_token_file.
+1. The authentication preference should always be given to the authentication token. If it's not passed, the path to domino token file
+takes precedence, otherwise the API key is used. If none of these three parameters are passed, then preference will be given to the
+domino token file from the corresponding environment variable, then to the API key from the corresponding environment variable.
```
dmitriyvolk, line 59: 👍

dmitriyvolk (approved):

tests/test_basic_auth.py
```diff
@@ -124,6 +124,21 @@ def test_auth_token_precedence():
...
+    """
+    Confirm that api key takes precedence over both token file and API key authentication
+    if they are not passed as parameters.
+    """
+    dummy_host = "http://domino.somefakecompany.com"
+    dummy_api_key = "top_secret_api_key"
+
+    d = Domino(host=dummy_host, project="anyuser/quick-start", api_key=dummy_api_key)
+    assert isinstance(d.request_manager.auth, requests.auth.HTTPBasicAuth), \
+        "With only api key passed as a parameter, it takes precedence over DOMINO_TOKEN_FILE_KEY_NAME"
```
dknupp, line 139: This is an exact duplicate of `test_object_creation_with_api_key` above.
```suggestion
```

dknupp: If you really want to be thorough, these are tests that you could consider adding.

- set a value for `DOMINO_USER_API_KEY_KEY_NAME` in the environment, also pass a different API key as a parameter, and confirm that the parameter is used
- set keys in the os.environ for both `DOMINO_TOKEN_FILE_KEY_NAME` and `DOMINO_USER_API_KEY_KEY_NAME`, and confirm that `DOMINO_TOKEN_FILE_KEY_NAME` gets used when no explicit auth is passed
- set a key in the os.environ for `DOMINO_USER_API_KEY_KEY_NAME` only, and confirm that it gets used when no other explicit auth is passed

domino/domino.py
```diff
@@ -73,7 +70,9 @@ def authenticate(self, api_key=None, auth_token=None, domino_token_file=None):
         """
         self.request_manager = _HttpRequestManager(get_auth_by_type(api_key=api_key,
                                                                     auth_token=auth_token,
-                                                                    domino_token_file=domino_token_file))
+                                                                    domino_token_file=domino_token_file,
+                                                                    api_key_from_env=os.getenv(DOMINO_USER_API_KEY_KEY_NAME),
+                                                                    domino_token_file_from_env=os.getenv(DOMINO_TOKEN_FILE_KEY_NAME)))
```
dknupp, line 75: I would revert this change, and handle all of the logic in `get_auth_by_type`. 
```suggestion
                                                                    domino_token_file=domino_token_file,
                                                                    api_key_from_env=os.getenv(DOMINO_USER_API_KEY_KEY_NAME))
```

domino/authentication.py
```diff
@@ -26,24 +26,29 @@ def __call__(self, r):
...
 
     if auth_token is not None:
         return BearerAuth(auth_token=auth_token)
     elif domino_token_file is not None:
         return BearerAuth(domino_token_file=domino_token_file)
-    else:
+    elif api_key is not None:
         return HTTPBasicAuth('', api_key)
+    else:
+        return self.get_auth_by_type(api_key=api_key_from_env, domino_token_file=domino_token_file_from_env)
```
dknupp, line 53: I feel like have a parameter called `something_from_env` is a code smell. Generally speaking, if something is in the environment, it doesn't have to be pass as a parameter. The fact that this method then somewhat unusually _called itself_ a second time under certain conditions made me think that this process should be simpler.

My suggestion is to simply encode the precedence all within this method. This seems like the most straightforward approach.
```suggestion
    if any([auth_token, domino_token_file, api_key]):
        if auth_token:
            return BearerAuth(auth_token=auth_token)
        elif domino_token_file:
            return BearerAuth(domino_token_file=domino_token_file)
        else:
            return HTTPBasicAuth('', api_key)
    else:
        # In the case that no authentication type was passed when this method
        # called, fall back to deriving the auth info from the environment.

        if os.getenv(DOMINO_TOKEN_FILE_KEY_NAME):
            return BearerAuth(domino_token_file=os.getenv(DOMINO_TOKEN_FILE_KEY_NAME))
        elif os.getenv(DOMINO_USER_API_KEY_KEY_NAME):
            return HTTPBasicAuth('', os.getenv(DOMINO_USER_API_KEY_KEY_NAME))
        else:
            # All attempts failed -- nothing to do but raise an error.
            raise RuntimeError("Unable to authenticate: no authentication provided")
```

domino/authentication.py
```diff
@@ -26,24 +26,29 @@ def __call__(self, r):
         return r
 
 
-def get_auth_by_type(api_key=None, auth_token=None, domino_token_file=None):
+def get_auth_by_type(api_key=None, auth_token=None, domino_token_file=None, api_key_from_env=None, domino_token_file_from_env=None):
```
dknupp, line 29: Revert this change.
```suggestion
def get_auth_by_type(api_key=None, auth_token=None, domino_token_file=None):
```

domino/authentication.py
dknupp: Note that you will need to change the first unit test as a result of this change.
```python
    with pytest.raises(AssertionError):
        Domino(host=dummy_host, project="anyuser/quick-start")
```
...will need to be
```python
    with pytest.raises(RuntimeError):
        Domino(host=dummy_host, project="anyuser/quick-start")
```

tests/test_basic_auth.py
ddl-alexpanin: > This is an exact duplicate of `test_object_creation_with_api_key` above.

It isn't. The new test assumes there is a token file found in the env, that's the test actually testing the changes :) Let me rename the test method to make it clearer.

ddl-alexpanin:
Thanks @dknupp! I incorporated most of your comments.

PR #126, opened 2022-05-16
[RM-1171] - added pypi link, and updated gitignore with pipfiles

ddl-olsonJD:
## Jira Links
https://dominodatalab.atlassian.net/browse/RM-1171

## Description
- Updated readme with pypi release link
- updated gitignore to ingore Pipfile* env files

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # python-domino
+[![PyPI version](https://badge.fury.io/py/dominodatalab.svg)](https://badge.fury.io/py/dominodatalab)
 
 This library provides bindings for the [Domino Data Lab API](https://dominodatalab.github.io/api-docs/).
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,4 +1,5 @@
 # python-domino
+[![PyPI version](https://badge.fury.io/py/dominodatalab.svg)](https://badge.fury.io/py/dominodatalab)
```
ddl-alexpanin, line 2: 1. Why can't you just use https://pypi.org/project/dominodatalab?
2. Why this link is needed? PyPI is a very standard thing so everyone should be able to find a package there knowing only its name. Note if you add the link, then the next release of dominodatalab in PyPI (e.g. 1.0.9) will also have the link on the project page in PyPI (as README.md is shared)

PR #50, opened 2020-01-09
Datasets list

katieshakman:
Cleaned up comments in the example get_datasets.py.

PR #136, opened 2022-06-22
[DOM-38906] - update datasets endpoints

ddl-olsonJD:
### Link to JIRA

[DOM-38906](https://dominodatalab.atlassian.net/browse/DOM-38906)

Domino had api endpoints available for dataset access, however, those endpoints were not usable via python API. Some updates were made to the datasets endpoints, which are also reflected in this PR.

This PR's goal is to complete the work that had started a while back to bring dataset access via python-domino, and is being updated.

### What is the solution?
Updated PR code base to reflect current version of [python-domino](https://github.com/dominodatalab/python-domino) and [domino](https://github.com/cerebrotech/domino). 

### Testing
Added Unit tests, as well as example test file file
- [x] Unit test(s)

### Pull Request Reminders

### References (optional)
- https://github.com/dominodatalab/python-domino/pull/50


ddl-olsonJD:
image

README.md
```diff
@@ -247,6 +247,70 @@ parameter in `job_start` method
 
 <hr>
 
+## Datasets
+
```
ddl-awroblicky, line 251: more just a nice to have, but maybe it could be helpful to define what a Dataset is.

I think the Domino documentation defines it as "A Domino Dataset is a collection of files that are available in user executions as a filesystem directory" (https://docs.dominodatalab.com/en/latest/user_guide/0a8d11/datasets-overview/)

domino/domino.py
```diff
@@ -423,7 +427,7 @@ def validate_spark_executor_count(executor_count, max_executor_count):
         def get_default_spark_settings():
             self.log.debug("Getting default spark settings")
             default_spark_setting_url = self._routes.default_spark_setting(
-                self._project_id
+                self.project_id
```
ddl-awroblicky, line 430: What are the tradeoffs involved in making this no longer private?

domino/domino.py
```diff
@@ -925,6 +929,74 @@ def model_version_publish(
         response = self.request_manager.post(url, json=request)
         return response.json()
 
+    # Dataset Functions
+    def datasets_list(self, project_id=None):
+        self.requires_at_least("3.6.0")
```
ddl-awroblicky, line 934: "3.6.0" seems to be used in several places, perhaps it could be a constant?

README.md
ddl-olsonJD: Agreed.

domino/domino.py
ddl-olsonJD: In addition to using it internally (as in the example above), it can also be used externally, particularly within datasets to filter current project datasets. It can also be used, can cases where someone may want to switch between projects, in ensuring current project.

ddl-olsonJD: Keeping it private would mean that users would be force to use a private property to reach the same goal.

domino/exceptions.py
```diff
@@ -4,6 +4,18 @@ class DominoException(Exception):
     pass
 
 
+class DatasetNotFoundException(DominoException):
+    """Run Not Found Exception"""
```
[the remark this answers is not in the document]
ddl-olsonJD: Thank you for catching this.

domino/domino.py
ddl-olsonJD: As 3.6.x is the current minimum version available, that will be set as min required, until 3.6 is fully deprecated.

PR #143, opened 2022-08-17
[DOM-40132] - preparing for release

ddl-olsonJD:
### Link to JIRA

[DOM-DOM-40132](https://dominodatalab.atlassian.net/browse/DOM-40132)

resolves issues around dependencies and missing documentations

### What is the solution?
Update dependencies with pinned version as well as updated documentation around dependency usage.
Also added tags documentation

### Testing

run installations steps insuring no dependency issues.

- [ ] Unit test(s)

### Pull Request Reminders


README.md
```diff
@@ -316,6 +316,44 @@ Note: datasets are first marked for deletion, then are deleted after a certain a
...
+
+provides detail information about the tag
+
+- _tag_id:_ id that identifies the specific project to be used
+
+### tags_add(tags, *project_id)
+
+Creates a tag, if it does not exist, and add it to the current the project
+
+- _tags (list):_ name that identifies the specific tag to be used
```
ddl-awroblicky, line 340: In addition to the above, I see some examples where the data type is on the right of the `:` in the readme 

 _dataset_ids: (list\[string\])_ 

and other cases where it is not underscored

_runId:_ string

This is fine for now but maybe in the future it would be good to standardize

ddl-awroblicky: also should it be `names that identify the specific tags to be used` since it is a list

tests/test_projects.py
```diff
@@ -87,6 +87,7 @@ def test_upload_file_to_project(default_domino_client):
     assert response.status_code == 201
     assert response.json()["path"] == "test_file.py"
 
+
```
ddl-awroblicky, line 90: why the extra line?

tests/test_data_sources_nods.py
```diff
@@ -0,0 +1,19 @@
+import os
+
+import pytest
+
+
+@pytest.mark.skipif(
+    os.getenv("DATA_SDK") == "yes", reason="Testing failure when dependency is missing"
+)
+def test_data_sources_import_fails():  # noqa
```
ddl-awroblicky, line 9: out of curiosity, what was the reasoning for moving this to a different file?

ddl-awroblicky: also what is the `_nods` part in the filename referring to?

README.md
ddl-olsonJD: moving this section to documents team, and yes.

tests/test_data_sources_nods.py
ddl-olsonJD: this is a pre-work for testing purposes with tox to test for different environment values. ds :: dataset downloaded

ddl-olsonJD: So, one set of test test if the dataset is downloaded, and one if not.

tests/test_projects.py
ddl-olsonJD: that's python standard spacing between functions.

ddl-awroblicky: that makes sense to add, but it doesn't seem related to any of the other changes

I would maybe have a separate pull request for it or wait until that file needs to be changed

tests/test_data_sources_nods.py
ddl-awroblicky: so I better understand, why does it need to be moved to a different file to test for different environment values?

so is nods short for no data sets?

tests/test_projects.py
ddl-olsonJD: that's a black - pre-commit fix. if the file had particularly issues, once touched, blacks updates.

ddl-awroblicky: I see, ok that makes sense to keep then

tests/test_data_sources_nods.py
ddl-olsonJD: yes.

ddl-olsonJD: see tox.ini, other wise, it returns false warning.

ddl-awroblicky: getting rid of a misleading warning sounds good

ddl-awroblicky: It wasn't obvious to me what "nods" meant so I think it could make sense to rename to something like "test_no_data_sources.py"

CHANGELOG.md
```diff
@@ -17,7 +17,9 @@ All notable changes to the `python-domino` library will be documented in this fi
 * Fixed issues with import from domino_data
 * Updated nbconvert dependency version to 6.3.0
 * updated request user-agent to python-domino/{version}
+* Updated dependency management
 * Changed file handling for `files_upload`
+* removed requirements.txt in favor setup.py
```
domino-jade-liu, line 22: I think it's nice to have a requirements.txt in addition to setup.py to list all dependencies. Why do you want to remove it?

ddl-awroblicky (approved):

CHANGELOG.md
ddl-olsonJD: In order to have a singular source of dependencies, otherwise, we risk having one being changed, leaving the other unchanged, which can then lead to dependency issues.

Issue #33, opened 2018-01-18
Error when calling files_upload


andscoop:
When calling files_upload I receive a 400 error if the fp does not have a '/' in front of the name. 
ex.
'test.py'
vs
'/test.py'

Should be an easy fix. Will patch if I find time

marks:
@andscoop this does seem to trip folks up every once in a while. Do you think Domino should return a better error or simply append the preceding `/` for users?

@macrogers @johnjoo1 interested in your thoughts on this one too.

peterdcotton:
I get a 404 error. Do you see anything wrong with this usage...

`def upload_data(local_file,remote_file):
    f = open(local_file, 'rb')
    domino = Domino("peter.cotton/percipience_data",
                api_key='7DunjmBLAHBLAHgzF',
                host='http://app.dominodatalab.com')
    r = domino.files_upload(remote_file, f)
    if r.status_code == 201:
        print(":) Upload successful")
    else:
        print(r.status_code)
        print("!! Upload failed")`

ddl-olsonJD:
resolved in pr
- https://github.com/dominodatalab/python-domino/pull/144

PR #144, opened 2022-08-17
[DOM-39831] - added forward slash check for file upload

ddl-tanner-green:
### Link to JIRA

[DOM-39831](https://dominodatalab.atlassian.net/browse)

When a file that is being uploaded does not have a forward slash at the beginning it throws an error

### What is the solution?

Added a checker to make sure the file still gets uploaded either way.

### Testing

- [ ] Unit test(s)

### Pull Request Reminders

### References (optional)
- https://github.com/dominodatalab/python-domino/issues/33

### DIFF
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -667,6 +667,8 @@ class Domino:
         return self._get(url)
 
     def files_upload(self, path, file):
+        if path[0] != "/":
+            path = "/" + path
         url = self._routes.files_upload(path)
         return self._put_file(url, file)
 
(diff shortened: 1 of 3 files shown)


domino/domino.py
```diff
@@ -667,6 +667,8 @@ def files_list(self, commitId, path="/"):
         return self._get(url)
 
     def files_upload(self, path, file):
+        if path[0] != "/":
```
ddl-olsonJD, line 670: nit picking here, but can we use startswith instead?

ddl-olsonJD:
Can we also update the changelog

ddl-olsonJD (changes requested):

ddl-olsonJD (approved):

PR #152, opened 2022-09-15
[DOM-36836] Validate HWT IDs for compute cluster properties when starting a job

mildmelon:
### Link to JIRA

[DOM-36836](https://dominodatalab.atlassian.net/browse/DOM-36836)

The parameter `compute_cluster_properties` in the `domino.job_start()` method contains HardwareTier IDs at `masterHardwareTierId` and `workerHardwareTierId` that are not validated before sending an HTTP request. See below for an example showing how this method could contain invalid HardwareTier IDs. The result of executing this code does not indicate any initial error, it shows that the job request was sent to the server and happily moves on without error. But when the job request is received on the server it is immediately failed as the HardwareTier IDs do not exist.

```python
domino.job_start(
    command="main.py",
    hardware_tier_id="small-k8s",
    compute_cluster_properties={
        "clusterType": "Ray",
        "computeEnvironmentId": "63222a311eeaac27c221432d",
        "computeEnvironmentRevisionSpec": "ActiveRevision",
        "masterHardwareTierId": "invalid_HWT_id",
        "workerHardwareTierId": "invalid_HWT_id",
        "workerCount": 1,
    }
)
```

### What is the solution?

The solution is to validate that the HardwareTier IDs in the `domino.job_start()` method parameter `compute_cluster_properties` are valid and exist. The validation is done on both `masterHardwareTierId` and `workerHardwareTierId`. This is done locally on the host computer as we only need to query the server for the list of existing HWT IDs and check to see if the desired HWT ID is present. I tested to ensure that valid HWT IDs still work and had a job run successfully.

### Testing

I created a new workspace on my dev-v2 environment and then pulled my branch for this repo. Then I installed the local project as a dependency `pip install -e .`. From there I used a Jupyter notebook and ran the script below, testing with both valid and invalid HWT IDs.

```python
import os
from domino import Domino

domino = Domino(
    "integration-test/quick-start",
    api_key=os.environ["DOMINO_USER_API_KEY"],
    host=os.environ["DOMINO_API_HOST"],
)

domino.job_start(
    command="main.py",
    hardware_tier_id="small-k8s",
    compute_cluster_properties={
        "clusterType": "Spark",
        "computeEnvironmentId": "63222a311eeaac27c221432d",
        "computeEnvironmentRevisionSpec": "ActiveRevision",
        "masterHardwareTierId": "small-kasd8s",
        "workerHardwareTierId": "small-k8s",
        "workerCount": 1,
    }
)
```

- [ ] Unit test(s)

### Pull Request Reminders


mildmelon:
I've also updated the version `1.1.1 -> 1.1.2` and added a new version section in the changelog.

hiiamelliott (approved):
LGTM– thanks Caidan.

domino/_version.py
```diff
@@ -1 +1 @@
-__version__ = "1.1.1"
+__version__ = "1.1.2"
```
ddl-alexpanin, line 1: Let's not do the version change yet. We already had 1.1.1 release (in github release and PyPI) for Domino 5.3.0. We can have 1.1.2 for 5.4.0.

ddl-alexpanin:
Also, https://dominodatalab.atlassian.net/browse/DOM-36836 has "backlog" as a fix version, what is the real fix version here? 5.4.0?

domino/domino.py
```diff
@@ -499,6 +499,12 @@ def throw_if_information_invalid(key: str, info: dict) -> bool:
                     f"'maxWorkerCount' is not supported in Domino {self._version}."
                 )
 
+            if "masterHardwareTierId" in compute_cluster_properties:
```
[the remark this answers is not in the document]
mildmelon: Thanks for the feedback, you're right that I don't need to check for the existence of the key. Although, after review it looks like we still need to check for the `masterHardwareTierId` as it can be optional when using an `MPI` cluster type.

See lines `450-460`:
https://github.com/dominodatalab/python-domino/blob/75ee9c47bd29894c4bac24056cad66387932f280/domino/domino.py#L443-L460

I've updated the code so that `workerHardwareTier` doesn't have the existence check.

domino/_version.py
mildmelon: Good to know, thanks for the tip. I've undo the version update and moved the changelog notes to the "Unreleased" section.

mildmelon:
> Also, https://dominodatalab.atlassian.net/browse/DOM-36836 has "backlog" as a fix version, what is the real fix version here? 5.4.0?

Yep, there is no specific version we need this released for as it doesn't rely on any specific changes from 5.4.0. I've updated the ticket with the fix version to be 5.4.0

ddl-alexpanin (approved):

ddl-olsonJD (approved):

domino/_version.py
ddl-olsonJD: @mildmelon version upgrades are done only at release time.

Issue #171, opened 2023-03-30
Add upload_file_to_dataset() method


ZeroCool2u:
Could we add an option to upload files to a Domino Dataset? The CLI is often inconvenient, because it doesn't support headless installations and we don't want to upload files to our code files directory, which is the only option right now via the Python package. There are unfortunately a number of situations in our corporate environment where 'pushing' to Domino is our only choice and we are unable to 'pull' the data via a regular scheduled job in Domino.

ddl-alexpanin:
@ZeroCool2u thanks, I opened a field request for the Data team in our internal JIRA.

ZeroCool2u:
Thank you!

ddl-ahmet-gyger:
Thanks for the feedback @ZeroCool2u - can you share more information on your use case?

ZeroCool2u:
@ddl-ahmet-gyger I can't share more information publicly. Can you sync up with our DDL CSM and Account Exec internally? Our CSM is Tony M and our Account Exec is Chris E. Hopefully you can narrow it down to who they are internally and my workplace, I just don't want to dox them by posting their full names in public. I just briefed them both on this FR in more detail this morning. Feel free to have them schedule some time to chat about this more with me and my team if you'd like as well.

ZeroCool2u:
Closed: Resolved by #185

PR #185, opened 2024-01-03
datasetRw: add dataset upload file capabilities to domino-python (multi-part, multi-thread)

ddl-giuliocapolino:
### Link to JIRA

[DOM-52840](https://dominodatalab.atlassian.net/browse)

Needed a new endpoint to handle dataset upload

### What is the solution?

Add endpoint as well as full client implementation + tests for single file, multi-part, multi-thread uploads

### Testing

- [x] Unit test(s)

### Pull Request Reminders

### DIFF
--- a/domino/datasets.py
+++ b/domino/datasets.py
@@ -73,30 +73,22 @@ class Uploader:
         self.upload_key = self.request_manager.post(start_upload_url, json=start_upload_body).json()
         if not self.upload_key:
             raise RuntimeError(f"upload key for {self.dataset_id} not found. Session could not start.")
-        # starting uploads
-        try:
-            paths = self._upload()
-            return paths
-        except Exception as e:
-            self.log.error("Upload failed. See error for details.")
-            raise e
+        return self._upload()
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         # catching errors
-        if not self.upload_key:
-            raise RuntimeError(f"upload key for {self.dataset_id} not found. Could not end session.")
-        if exc_type != None:
+        if exc_type is not None:
             self.log.error(f"Upload for dataset {self.dataset_id} and file {self.local_path_file_or_directory} failed, "
                            f"attempting to cancel session. Please try again.")
             self.log.error(f"Error type: {exc_val}. Error message: {exc_tb}.")
             if not isinstance(exc_type, ValueError):
-                self._cancel_upload_session() # is it is a ValueError, canceling session would fail
+                self._cancel_upload_session()  # is it is a ValueError, canceling session would fail
             return
-        # ending snapshot uplaod
+        # ending snapshot upload
         try:
             url = self.routes.datasets_end_upload(self.dataset_id, self.upload_key, self.target_relative_path)
             self.request_manager.get(url)
-            self.log.info("Upload session completed successfully.")
+            self.log.info("Upload session ended successfully.")
         except Exception as e:
             self.log.error("Ending snapshot upload failed. See error for details. Attempting to cancel "
                            "upload session.")
@@ -104,16 +96,12 @@ class Uploader:
 
     def _cancel_upload_session(self):
         url = self.routes.datasets_cancel_upload(self.dataset_id, self.upload_key)
-        try:
-            self.request_manager.get(url)
-            self.log.info("Upload session cancelled successfully.")
-        except Exception as e:
-            self.log.info("Cancelling upload session failed. See error for details.")
-            raise e
+        self.request_manager.get(url)
+        self.log.info("Upload session cancelled successfully.")
 
     def _create_chunk_queue(self) -> list[UploadChunk]:
         if not os.path.exists(self.local_path_file_or_directory):
-            raise ValueError(f"Path {self.local_path_file_or_directory} does not exist.")
+            raise ValueError(f"local file or directory {self.local_path_file_or_directory} does not exist.")
         if os.path.isfile(self.local_path_file_or_directory):
             return self._create_chunks(self.local_path_file_or_directory)
         chunk_q = []
@@ -143,7 +131,7 @@ class Uploader:
         with ThreadPoolExecutor(self.max_workers) as executor:
             # list ensures all the threads are complete before returning results
             results = list(executor.map(self._upload_chunk, q))
-        return (self.local_path_file_or_directory)
+        return self.local_path_file_or_directory
 
     def _upload_chunk(self, chunk: UploadChunk) -> None:
         # read the file chunk
--- a/domino/datasets.py
+++ b/domino/datasets.py
@@ -73,26 +73,38 @@ class Uploader:
         self.upload_key = self.request_manager.post(start_upload_url, json=start_upload_body).json()
         if not self.upload_key:
             raise RuntimeError(f"upload key for {self.dataset_id} not found. Session could not start.")
-        return self._upload()
+        return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         # catching errors
         if exc_type is not None:
-            self.log.error(f"Upload for dataset {self.dataset_id} and file {self.local_path_file_or_directory} failed, "
-                           f"attempting to cancel session. Please try again.")
+            self.log.error(f"Upload for dataset {self.dataset_id} and file or directory "
+                           f"`{self.local_path_file_or_directory}` failed, attempting to cancel session. "
+                           f"Please try again.")
             self.log.error(f"Error type: {exc_val}. Error message: {exc_tb}.")
             if not isinstance(exc_type, ValueError):
                 self._cancel_upload_session()  # is it is a ValueError, canceling session would fail
-            return
+            return False
         # ending snapshot upload
         try:
             url = self.routes.datasets_end_upload(self.dataset_id, self.upload_key, self.target_relative_path)
             self.request_manager.get(url)
             self.log.info("Upload session ended successfully.")
-        except Exception as e:
+            return True
+        except:
             self.log.error("Ending snapshot upload failed. See error for details. Attempting to cancel "
                            "upload session.")
             self._cancel_upload_session()
+            return False
+
+    def upload(self):
+        if not self.upload_key:
+            raise RuntimeError(f"upload key for {self.dataset_id} not found. Please start session before uploading.")
+        q = self._create_chunk_queue()
+        with ThreadPoolExecutor(self.max_workers) as executor:
+            # list ensures all the threads are complete before returning results
+            results = list(executor.map(self._upload_chunk, q))
+        return self.local_path_file_or_directory
 
     def _cancel_upload_session(self):
         url = self.routes.datasets_cancel_upload(self.dataset_id, self.upload_key)
@@ -123,15 +135,6 @@ class Uploader:
                             target_chunk_size=self.target_chunk_size, total_chunks=total_chunks,
                             upload_key=self.upload_key)
                 for chunk_num in range(starting_index, total_chunks + 1)]
-
-    def _upload(self):
-        if not self.upload_key:
-            raise RuntimeError(f"upload key for {self.dataset_id} not found. Please start session before uploading.")
-        q = self._create_chunk_queue()
-        with ThreadPoolExecutor(self.max_workers) as executor:
-            # list ensures all the threads are complete before returning results
-            results = list(executor.map(self._upload_chunk, q))
-        return self.local_path_file_or_directory
 
     def _upload_chunk(self, chunk: UploadChunk) -> None:
         # read the file chunk
--- a/domino/datasets.py
+++ b/domino/datasets.py
@@ -1,0 +1,169 @@
+import hashlib
+import math
+import os
+import time
+from concurrent.futures import ThreadPoolExecutor
+from logging import Logger
+from typing import AnyStr
+
+from domino.http_request_manager import _HttpRequestManager
+from domino.routes import _Routes
+
+MAX_WORKERS = 10
+MAX_UPLOAD_ATTEMPTS = 10
+MB = 2 ** 20  # 2^20 bytes - 1 Megabyte
+SLEEP_TIME_IN_SEC = 5
+UPLOAD_READ_TIMEOUT_IN_SEC = 30
+
+
+class UploadChunk:
+    def __init__(
+            self,
+            absolute_path: str,
+            chunk_number: int,
+            dataset_id: str,
+            file_name: str,
+            file_size: int,
+            upload_key: str,
+            identifier: str,
+            relative_path: str,
+            target_chunk_size: int,
+            total_chunks: int,
+    ):
+        self.absolute_path = absolute_path
+        self.chunk_number = chunk_number
+        self.dataset_id = dataset_id
+        self.file_name = file_name
+        self.file_size = file_size
+        self.identifier = identifier
+        self.relative_path = relative_path
+        self.target_chunk_size = target_chunk_size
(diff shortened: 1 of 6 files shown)


domino/domino.py
```diff
@@ -1047,6 +1050,94 @@ def datasets_remove(self, dataset_ids: list):
...
+                upload_chunk_url = self._routes.datasets_upload_chunk(dataset_id, upload_key, chunk_number, total_chunks, chunk_size, len(chunk_data), identifier, chunk_relative_path, chunk_checksum)
+                upload_chunk_response = self._post(upload_chunk_url, chunk_data)  # chunk data is likely wrong here
+                # add check upload chunk response is successful
+                self.log.info(f"Uploaded chunk {chunk_number} of {total_chunks} for {file_name}")
+                if upload_try > MAX_UPLOAD_ATTEMPTS:
+                    raise RuntimeError(f"Uploading chunk {chunk_number} of {total_chunks} for {file_name} failed. Please try again")
+
+        else:
+            self.log.info(f"Skipping chunk {chunk_number} of {total_chunks} for {file_name}")
+
```
ddl-giuliocapolino, line 1140: missing another endpoint for end_upload

domino/datasets.py
```diff
@@ -0,0 +1,174 @@
...
+
+FILE_UPLOAD_SETTING_DEFAULT = "Overwrite"
+MAX_WORKERS = 10
+MAX_UPLOAD_ATTEMPTS = 10
+MB = 2 ** 20  # 2^20 bytes - 1 Megabyte
+SLEEP_TIME_IN_SEC = 5
+UPLOAD_READ_TIMEOUT_IN_SEC = 30
+
+
+class UploadChunk:
```
ddl-gabrielhaim, line 20: seems to be a good use case for a dataclass https://docs.python.org/3/library/dataclasses.html

domino/datasets.py
```diff
@@ -0,0 +1,190 @@
...
+        url = self.routes.datasets_cancel_upload(self.dataset_id, self.upload_key)
+        return self.request_manager.get(url)
+
+    def end_upload_session(self):
+        if not self.upload_key:
+            raise RuntimeError(f"upload key for {self.dataset_id} not found. Could not end session.")
+        url = self.routes.datasets_end_upload(self.dataset_id, self.upload_key, self.target_relative_path)
+        return self.request_manager.get(url)
+
+    def _create_chunk_queue(self) -> [UploadChunk]:
```
ddl-gabrielhaim, line 103: I think you have to use `list[UploadChunk]` for proper type hinting here

domino/domino.py
```diff
@@ -1047,6 +1047,50 @@ def datasets_remove(self, dataset_ids: list):
...
+            routes=self._routes,
+
+            file_upload_setting=file_upload_setting,
+            max_workers=max_workers,
+            target_chunk_size=target_chunk_size,
+            target_relative_path=target_relative_path
+        )
+
+        try:
+            uploader.start_upload_session()
```
ddl-gabrielhaim, line 1085: I feel like you could easily make this a context manager https://www.pythonmorsels.com/creating-a-context-manager/ to avoid having the user of this class know which method to call first (since there are no arguments once created)

domino/datasets.py
```diff
@@ -0,0 +1,190 @@
...
+        if os.path.isfile(self.local_path_file_or_directory):
+            return self._create_chunk(self.local_path_file_or_directory)
+        else:
```
ddl-gabrielhaim, line 106: python like early return, no need for `else` here

domino/datasets.py
```diff
@@ -0,0 +1,190 @@
...
+            nested_q = []
+            for dirpath, dirnames, filenames in os.walk(self.local_path_file_or_directory):
+                for filename in filenames:
+                    # construct the relative path for each file
+                    relative_path_to_file = os.path.join(dirpath, filename)
+                    # append chunk to queue
+                    nested_q.append(self._create_chunk(relative_path_to_file))
+            # flattening list of chunks
+            return [chunk for chunks in nested_q for chunk in chunks]
```
ddl-gabrielhaim, line 115: WAT is happening here 😆 . Can you name things differently? is there a way to flatten this in the previous for loop?

domino/datasets.py
```diff
@@ -0,0 +1,190 @@
...
+            chunk_data = file.read(chunk.target_chunk_size)
+
+        # testing chunk
+        should_upload, checksum = self._test_chunk(chunk, chunk_data)
+
+        # uploading chunk
+        if should_upload:
+            upload_try = 1
+            uploaded = False
+            while upload_try <= MAX_UPLOAD_ATTEMPTS and not uploaded:
```
ddl-gabrielhaim, line 144: big fan of making good use a decorator for this kind of logic https://pypi.org/project/retry/. This allow clearer code by making the function do only what it is supposed to do

domino/datasets.py
```diff
@@ -0,0 +1,190 @@
...
+        self.file_name = file_name
+        self.file_size = file_size
+        self.identifier = identifier
+        self.relative_path = relative_path
+        self.target_chunk_size = target_chunk_size
+        self.total_chunks = total_chunks
+        self.upload_key = upload_key
+
+
+class Uploader:
```
ddl-gabrielhaim, line 46: see my comment about making this a context manager class

tests/test_datasets.py
```diff
@@ -94,6 +94,20 @@ def test_datasets_details(default_domino_client, random_seq):
...
+def test_datasets_upload(default_domino_client):
+    datasets_id = default_domino_client.datasets_ids(default_domino_client.project_id)[
+        1
+    ]
+    local_path_to_file = "test_datasets.py"
+    response = default_domino_client.datasets_upload_file(datasets_id, local_path_to_file)
+
+    assert response.status_code == 200
+    assert "filecache" in response.json()
+q
```
ddl-eric-jin, line 109: Is this `q` supposed to be here?

domino/datasets.py
```diff
@@ -0,0 +1,190 @@
...
+                    end_time_ns = time.time_ns()
+                    duration_ns = end_time_ns - start_time_ns
+                    bandwidth_bytes_per_second = actual_chunk_size / duration_ns * 1000000000.0
+                    self.log.info(f"Uploaded chunk {chunk.chunk_number} of {chunk.total_chunks} for {chunk.file_name} "
+                                  f"in {duration_ns / 1_000_000:.1f}ms ({bandwidth_bytes_per_second:.1f} B/s)")
+                    uploaded = True
+                except Exception:
+                    if upload_try > MAX_UPLOAD_ATTEMPTS:
+                        raise Exception(f"Uploading chunk {chunk.chunk_number} of {chunk.total_chunks} "
+                                           f"for {chunk.file_name} failed. Please try again")
```
ddl-eric-jin, line 171: Nit: Can add period after "Please try again"

domino/routes.py
```diff
@@ -203,6 +206,54 @@ def datasets_create(self):
...
+            f"&resumableTotalChunks={total_chunks}&checksum={quote(checksum)}"
+        )
+
+    def datasets_cancel_upload(self, dataset_id, upload_key):
+        return self.host + f"/v4/datasetrw/datasets/{dataset_id}/snapshot/file/cancel/{upload_key}"
+
+    def datasets_end_upload(self, dataset_id, upload_key, target_relative_path=None):
+        url = self.host + f"/v4/datasetrw/datasets/{dataset_id}/snapshot/file/end/{upload_key}"
+        if target_relative_path:
+            url += f"?targetRelativePath={target_relative_path}"
```
ddl-eric-jin, line 254: Nice!

domino/datasets.py
```diff
@@ -0,0 +1,190 @@
...
+        self.absolute_path = absolute_path
+        self.chunk_number = chunk_number
+        self.dataset_id = dataset_id
+        self.file_name = file_name
+        self.file_size = file_size
+        self.identifier = identifier
+        self.relative_path = relative_path
+        self.target_chunk_size = target_chunk_size
+        self.total_chunks = total_chunks
+        self.upload_key = upload_key
```
ddl-eric-jin, line 43: Small Nit: Might be clearer to match the order or the arguments of `__init__` with the order the variables are assigned.

domino/datasets.py
ddl-eric-jin, line 171: Nit: Could add period at the end of message.

tests/test_datasets.py
ddl-giuliocapolino: haha no!! thanks :)

domino/datasets.py
```diff
@@ -0,0 +1,197 @@
...
+            chunk_number: int,
+            dataset_id: str,
+            file_name: str,
+            file_size: int,
+            identifier: str,
+            relative_path: str,
+            target_chunk_size: int,
+            total_chunks: int,
+            upload_key: str,
+    ):
```
ddl-gabrielhaim, line 49: you don't need the `__init__` in a dataclass (that was the goal of my comment)

domino/datasets.py
```diff
@@ -0,0 +1,197 @@
...
+            "filePaths": [],
+            "fileCollisionSetting": self.file_upload_setting
+        }
+        start_upload_url = self.routes.datasets_start_upload(self.dataset_id)
+        self.upload_key = self.request_manager.post(start_upload_url, json=start_upload_body).json()
+        if not self.upload_key:
+            raise RuntimeError(f"upload key for {self.dataset_id} not found. Session could not start.")
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
```
ddl-gabrielhaim, line 101: you might want to check on the exc_type to cancel the upload session in the event an exception was raised? This way you handle the whole lifecycle of the upload here instead of leaving that to the user

domino/domino.py
```diff
@@ -1047,6 +1048,51 @@ def datasets_remove(self, dataset_ids: list):
...
+        ) as uploader:
+            try:
+                response = uploader.upload()
+                return response
+            except ValueError:
+                self.log.error(f"Erroneous input parameter. Please fix the parameters and try again")
+            except Exception as e:
+                self.log.error(f"Upload for dataset {dataset_id} and file {local_path_to_file_or_directory} failed, "
+                               f"canceling session. Please try again.")
+                uploader.cancel_upload_session()
```
ddl-gabrielhaim, line 1092: see previous comment, want to avoid that here

domino/datasets.py
ddl-giuliocapolino: ahh i see!

domino/datasets.py
```diff
@@ -0,0 +1,198 @@
...
+        self.upload_key = self.request_manager.post(start_upload_url, json=start_upload_body).json()
+        if not self.upload_key:
+            raise RuntimeError(f"upload key for {self.dataset_id} not found. Session could not start.")
+        # starting uploads
+        try:
+            paths = self._upload()
+            return paths
+        except Exception as e:
+            self.log.error("Upload failed. See error for details.")
+            raise e
```
ddl-gabrielhaim, line 82: I don't think you need to use `try..catch` here since it gets captured in the __exit__

domino/datasets.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        # catching errors
+        if not self.upload_key:
+            raise RuntimeError(f"upload key for {self.dataset_id} not found. Could not end session.")
```
ddl-gabrielhaim, line 87: you might want to check the exception before this since an error can happen before the upload_key is set

domino/datasets.py
```diff
@@ -0,0 +1,198 @@
...
+        if exc_type != None:
```
ddl-gabrielhaim, line 88: `if exc_type is not None` is the python way of checking null values

domino/datasets.py
```diff
@@ -0,0 +1,198 @@
...
+            self.log.error(f"Upload for dataset {self.dataset_id} and file {self.local_path_file_or_directory} failed, "
+                           f"attempting to cancel session. Please try again.")
+            self.log.error(f"Error type: {exc_val}. Error message: {exc_tb}.")
+            if not isinstance(exc_type, ValueError):
+                self._cancel_upload_session() # is it is a ValueError, canceling session would fail
```
ddl-gabrielhaim, line 93: what happen if this raise an exception? afaik `__exit__` must return True or False https://docs.python.org/3/library/contextlib.html#contextlib.ContextDecorator

domino/datasets.py
```diff
@@ -0,0 +1,198 @@
...
+            self._cancel_upload_session()
+
+    def _cancel_upload_session(self):
+        url = self.routes.datasets_cancel_upload(self.dataset_id, self.upload_key)
+        try:
+            self.request_manager.get(url)
+            self.log.info("Upload session cancelled successfully.")
+        except Exception as e:
+            self.log.info("Cancelling upload session failed. See error for details.")
+            raise e
```
ddl-gabrielhaim, line 112: probably want to prevent the exception from being re-raise (see previous comment)

domino/domino.py
```diff
@@ -1047,6 +1048,44 @@ def datasets_remove(self, dataset_ids: list):
...
+            log=self.log,
+            request_manager=self.request_manager,
+            routes=self._routes,
+
+            file_upload_setting=file_upload_setting,
+            max_workers=max_workers,
+            target_chunk_size=target_chunk_size,
+            target_relative_path=target_relative_path
+        ) as path:
+            if path:
```
ddl-gabrielhaim, line 1084: in what scenario is path not evaluating to true?

domino/datasets.py
ddl-giuliocapolino: it raises it! modifying below.

domino/domino.py
ddl-giuliocapolino: ah previously I was logging “Upload complete” here and was trying to find a way around the possibility of the ending failing - this is a remnant of that. good catch!

domino/datasets.py
```diff
@@ -0,0 +1,186 @@
...
+        # creating upload session
+        start_upload_body = {
+            "filePaths": [],
+            "fileCollisionSetting": self.file_upload_setting
+        }
+        start_upload_url = self.routes.datasets_start_upload(self.dataset_id)
+        self.upload_key = self.request_manager.post(start_upload_url, json=start_upload_body).json()
+        if not self.upload_key:
+            raise RuntimeError(f"upload key for {self.dataset_id} not found. Session could not start.")
+        return self._upload()
```
ddl-gabrielhaim, line 76: not sure whether you should let the user (user of the context manager) initiate the upload or not, seems more correct to let them do it in case they want to do some logging before and after.

domino/datasets.py
```diff
@@ -0,0 +1,186 @@
...
+            return
+        # ending snapshot upload
+        try:
+            url = self.routes.datasets_end_upload(self.dataset_id, self.upload_key, self.target_relative_path)
+            self.request_manager.get(url)
+            self.log.info("Upload session ended successfully.")
+        except Exception as e:
+            self.log.error("Ending snapshot upload failed. See error for details. Attempting to cancel "
+                           "upload session.")
+            self._cancel_upload_session()
```
ddl-gabrielhaim, line 95: see previous comment, the __exit__ method should return true/false

ddl-gabrielhaim (approved):
Approving but couple of comments that we might want to iterate on a future PR.

ZeroCool2u:
Big hype, this seems like it closes #171! Thank you!

PR #194, opened 2024-01-22
[DOM-52828] Add interface vals to the domino task

ddl-ryan-connor:
### Link to JIRA

https://dominodatalab.atlassian.net/browse/DOM-52824

Flyte input downloader init container and output uploader sidecar container need the interfaces for the input/output provided as args to their container commands in base64 string encoded format.

### What is the solution?

After instantiating the DominoTask, serialize the tasks input/output interfaces to base64 string and provide those as job config so they can be sent along to the job api by the flyte agent.

### Testing

Tested the serialized args produced from the python code with an arbitrary domino job that uses the init and sidecar containers, and placed data matching the interface at the expected input location in s3. Init sidecar downloaded the data, the code used the data and saved to the local output dir, then the sidecar successfully uploaded the output to s3.

### Pull Request Reminders

- [ ] Has the JIRA ticket(s) been linked above?

### DIFF
--- a/domino/flyte/task.py
+++ b/domino/flyte/task.py
@@ -82,11 +82,6 @@ class DominoJobConfig(object):
     ComputeClusterProperties: Optional[ClusterProperties] = None
     VolumeSizeGiB: Optional[float] = None
     ExternalVolumeMountIds: Optional[List[str]] = None
-    # we want to provide the following to the flyte agent, but these are set by the DominoJobTask constructor 
-    # and so would never be expected to be passed into a DominoJobConfig constructor by a user.
-    # they are included here for completeness documenting what task config values are sent to/read by the flyte agent.
-    _InputInterfaceBase64: Optional[str] = None
-    _InputOutputInterfaceBase64: Optional[str] = None
 
     def toJson(self) -> Dict[str, Any]:
         return {
@@ -165,8 +160,6 @@ class DominoJobTask(AsyncAgentExecutorMixin, PythonTask[DominoJobConfig]):
             resolved_job_config["apiKey"] = domino_job_config.ApiKey
             resolved_job_config["command"] = domino_job_config.Command
             resolved_job_config["title"] = domino_job_config.Title
-            resolved_job_config["inputInterfaceBase64"] = None
-            resolved_job_config["inputOutputInterfaceBase64"] = None
 
             return resolved_job_config            
         else:
--- a/domino/flyte/task.py
+++ b/domino/flyte/task.py
@@ -140,7 +140,7 @@ class DominoJobTask(AsyncAgentExecutorMixin, PythonTask[DominoJobConfig]):
             InputInterfaceBase64=base64.b64encode(serialized_input_interface),
             InputOutputInterfaceBase64=base64.b64encode(serialized_input_output_interface),
         )
-        domino_job_config.PipelineConfig = pipelineConfig
+        task.task_config.PipelineConfig = pipelineConfig
 
         return task
 
--- a/domino/flyte/task.py
+++ b/domino/flyte/task.py
@@ -5,6 +5,7 @@ from enum import Enum
 from typing import Any, Dict, Optional, Type, List
 from datetime import timedelta
 from dataclasses import dataclass, asdict
+import base64
 from flytekit.configuration import SerializationSettings
 from flytekit.core.base_task import PythonTask, TaskMetadata
 from flytekit.core.interface import Interface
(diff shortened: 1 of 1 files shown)


domino/flyte/task.py
```diff
@@ -149,6 +170,7 @@ def _resolve_job_properties(self, domino_job_config: DominoJobConfig) -> Dict[st
             resolved_job_config["apiKey"] = domino_job_config.ApiKey
             resolved_job_config["command"] = domino_job_config.Command
             resolved_job_config["title"] = domino_job_config.Title
+            resolved_job_config["pipelineConfig"] = asdict(PipelineConfig())
```
noahjax, line 173: Since we just return a dict with no required keys and you're just inserting an empty `PipelineConfig` its probably fine to omit this line altogether.

domino/flyte/task.py
```diff
@@ -112,16 +121,28 @@ def __init__(
...
+        # the flyte init container (which downloads inputs) and flyte sidecar (which uploads outputs) require these
+        # base64 string encodings of the input/output interfaces as args to their container startup commands
+        # TODO: can detect "empty interfaces" here and omit the interface args so dont even create the init/sidecar containers if not needed
+        serialized_input_interface = task._interface.to_flyte_idl().inputs.SerializeToString()
+        serialized_input_output_interface = job._interface.to_flyte_idl().SerializeToString()
+        pipelineConfig = PipelineConfig(
+            InputInterfaceBase64=base64.b64encode(serialized_input_interface),
+            InputOutputInterfaceBase64=base64.b64encode(serialized_input_output_interface),
+        )
+        task.task_config.PipelineConfig = pipelineConfig
```
noahjax, line 143: Seems like you could create the `Interface` you use here before you call `super.__init__()` and avoid having to update the task after it is created. Not sure if it matters either way.

Also, is there a reason that one of these uses `task._interface` and the other uses `job._interface`? I'm not sure where `job` is defined here.

ddl-ryan-connor: to the `job` question -- it was a copy/paste messup from my other code -- will change to `task`

domino/flyte/task.py
```diff
@@ -81,6 +82,11 @@ class DominoJobConfig(object):
     ComputeClusterProperties: Optional[ClusterProperties] = None
     VolumeSizeGiB: Optional[float] = None
     ExternalVolumeMountIds: Optional[List[str]] = None
+    # we want to provide the following to the flyte agent, but these are set by the DominoJobTask constructor 
+    # and so would never be expected to be passed into a DominoJobConfig constructor by a user.
+    # they are included here for completeness documenting what task config values are sent to/read by the flyte agent.
+    _InputInterfaceBase64: Optional[str] = None
```
ddl-ryan-connor, line 88: @noahjax this part is a bit odd, but including them here seemed better than leaving out -- what do you think?

noahjax: I'd leave them out, I don't think the user needs to know or think about these

domino/flyte/task.py
```diff
@@ -149,6 +165,8 @@ def _resolve_job_properties(self, domino_job_config: DominoJobConfig) -> Dict[st
             resolved_job_config["apiKey"] = domino_job_config.ApiKey
             resolved_job_config["command"] = domino_job_config.Command
             resolved_job_config["title"] = domino_job_config.Title
+            resolved_job_config["inputInterfaceBase64"] = None
+            resolved_job_config["inputOutputInterfaceBase64"] = None
```
noahjax, line 169: These lines are basically noops

domino/flyte/task.py
```diff
@@ -116,12 +124,20 @@ def __init__(
...
             outputs=outputs,
             metadata=TaskMetadata(retries=0, timeout=timedelta(hours=3)),
             **kwargs,
         )
+        # Interface class doesn't have to_flyte_idl() so can't seem to get the correct base64 vals before instantiating the task
+        # the flyte init container (which downloads inputs) and flyte sidecar (which uploads outputs) require these
+        # base64 string encodings of the input/output interfaces as args to their container startup commands
+        # TODO: can detect "empty interfaces" here and omit the interface args so dont even create the init/sidecar containers if not needed
+        serialized_input_interface = self._interface.to_flyte_idl().inputs.SerializeToString()
+        serialized_input_output_interface = self._interface.to_flyte_idl().SerializeToString()
```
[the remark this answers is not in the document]
ddl-ryan-connor: there are different: one is the input interface, and one is both the input output interface.

this is how the flyte containers expect the args.

domino/flyte/task.py
ddl-ryan-connor: ok, i'll take them out

domino/flyte/task.py
ddl-ryan-connor: sure but i think explicit is almost always better than implicit. but i guess since im taking the args out of the domino config object i can take this out

domino/flyte/task.py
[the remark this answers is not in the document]
noahjax: Nvm, am illiterate

noahjax (approved):

PR #196, opened 2024-04-16
[DOM-54635] - updated requests session with retry

ddl-olsonJD:
### Link to JIRA

[DOM-54635](https://dominodatalab.atlassian.net/browse/DOM-54635)

Some long-running jobs are failing with ConnectionError due to connections being dropped.

### What is the solution?
We updated the request session with configurable retries to support connection timeouts, preventing long-running jobs from failing due to connection drops or long timeouts.

### Testing
- [x] Unit test(s)

### Pull Request Reminders

### DIFF
--- a/domino/http_request_manager.py
+++ b/domino/http_request_manager.py
@@ -4,11 +4,15 @@ from http import HTTPStatus
 
 import requests
 from bs4 import BeautifulSoup
+from requests.adapters import HTTPAdapter, Retry
 from requests.auth import AuthBase
 
 from ._version import __version__
 from .constants import DOMINO_VERIFY_CERTIFICATE
 from .exceptions import ReloginRequiredException
+
+
+DOMINO_API_RETRIES = 5
 
 class _SessionInitializer:
     def __initialize__(self, session):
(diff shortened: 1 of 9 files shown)


ddl-olsonJD:
image

ddl-olsonJD:
Some tests are failing and being ticketed.

domino/http_request_manager.py
```diff
@@ -23,17 +27,31 @@ class _HttpRequestManager:
...
+            self.auth.__initialize__(self._set_session())
+
+    def _set_session(self):
+        """
+        Initialize a request session with retry to help manage connections drop.
+        """
+        self.session = requests.Session()
+        retries = Retry(
+            total=int(os.getenv("DOMINO_MAX_RETRIES", str(R_SESSION_MAX_RETRIES))),
+            backoff_factor=1,  # retry seconds
```
ddl-s-rodriguez, line 42: Is the backoff exponential?

tests/test_domino.py
```diff
@@ -16,3 +28,19 @@ def test_versioning(requests_mock, dummy_hostname):
     with pytest.raises(Exception):
         dom.requires_at_least("5.11.0")
 
+def test_request_session():
+     request_manager = _HttpRequestManager(auth=TestAuth())
+     start_time = time.time()
+     try:
+         response = request_manager.request_session.get(
+             'https://httpbin.org/delay/10',
```
ddl-s-rodriguez, line 36: Do we actually want to hit httpbin? Or should this. behavior be mocked?

ddl-s-rodriguez (approved):

domino/http_request_manager.py
```diff
@@ -23,17 +27,31 @@ class _HttpRequestManager:
...
+            status_forcelist=[502, 503, 504],
```
ddl-tobi-lichtig, line 43: A couple of quick checks: Does this also retry on the following?:
- 429 status codes, ideally awaiting the necessary time specified in the headers
- 408 status codes (not sure if we ever use them, but generally retrying on those makes sense)
- socket-level errors such as ECONNRESET and ECONNTIMEOUT

ddl-olsonJD: 408 does make sense, and we can include that, 
ECONNRESET and ECONNTIMEOUT and being handled, 
429, This can be handled by the end user.

tests/test_domino.py
ddl-olsonJD: Updated.

PR #199, opened 2024-04-22
[DOM-54068] - Updating for finops

ddl-olsonJD:
### Link to JIRA

[DOM-54068](https://dominodatalab.atlassian.net/browse/DOM-54068)

Billing tags are being used as part of cost governance to allow admins to track cost and resource usage within their organizations. Billing tags are being used during project creation in case billing tags are required. The current endpoints covered by Python Domino does not cover the usage of billing tags nor the creation of projects with billing tas

### What is the solution?

Cover budgets and billing tag endpoints to allow budgets and billing tag management through Python Domino, as well as updating project creation to use v4 which allows more fields to be configurable.

### Testing

Tested against live deployments, using unit test as well as examples. 

- [x] Unit test(s)

### Pull Request Reminders

- [x] Has the [changelog](https://github.com/dominodatalab/python-domino/blob/master/CHANGELOG.md) been updated


domino/routes.py
```diff
@@ -291,5 +289,31 @@ def find_project_by_owner_name_and_project_name_url(self):
         )
 
     # User URLs
-    def users_get(self):
+    def users_get(self) -> str:
         return self.host + "/v4/users"
+
+    # Budgets and Billing Tags
+    def budget_settings(self) -> str:
+        return self.host + "/v4/cost/budgets/global/alertsSettings"
```
ddl-mmahmoud, line 297: Is it possible to use the public endpoints instead of the v4 endpoints?

domino/routes.py
```diff
@@ -61,11 +59,11 @@ def runs_list(self):
     def runs_start(self):
         return self._build_project_url() + "/runs"
 
-    def runs_status(self, runId):
-        return self._build_project_url() + "/runs/" + runId
+    def runs_status(self, run_id):
+        return self._build_project_url() + "/runs/" + run_id
```
ddl-mmahmoud, line 63: Please don't change this as it could be called by clients as a keyword argument, meaning this is a breaking change
`dom.runs_status(runId="foo")`

ddl-mmahmoud: Or accept both and prefer `run_id`.

README.md
```diff
@@ -107,12 +107,206 @@ See
...
+Remove a billing tag from a specified project
+Requires Admin permission
+
+-   *billing_tag:* billing tag string to filter projects by
+
+-   *offset:* The index of the start of the page, where checkpointProjectId is index 0.
+If the offset is negative the project it points to will be the end of the page.
+-   *page_size:* The number of record to return per page.
+
+-   *checkpoint_project_id:* (Optional) project id that marks the start of where to look for the desired page from
```
ddl-mmahmoud, line 264: Let's not expose this parameter

README.md
```diff
@@ -107,12 +107,206 @@ See
...
+
+-   *name_filter:* matches projects by name substring
+
+-   *sort_by:* (Optional) field to sort the projects on
+
+-   *sort_order:* (Optional) Whether to sort in asc or desc order
+
+-   *missing_tag_only:* (Optional) determine whether to only return projects with missing tag
```
ddl-mmahmoud, line 272: Should we make a separate method to simply get all the projects without a billing tag?

README.md
```diff
@@ -107,12 +107,206 @@ See
...
+-   *notify_org_owner:* whether to notify organizations owners on projects reaching threshold.
+
+### budget_alerts_targets_update(targets)
+
+Update the current budget alerts settings with additional email targets per budget label
+Requires Admin permission
+
+-   *targets:* dictionary of budget labels and list of email addresses
+
+### active_billing_tags_list()
```
ddl-mmahmoud, line 187: How do you feel about `billing_tags_list_active`?  It would be kinda nice if all the billing tag methods started with `billing_tag`, although I don't think this is a big deal.

README.md
ddl-olsonJD: any particular reason? particularly since the user can get the list of projects with the ideas already?

README.md
ddl-olsonJD: that works

README.md
ddl-mmahmoud: The semantics of this parameter are pretty nonobvious.  Like is this project offset in a list sorted by project id, or is it project offset in the sort from the query?  If this is provided along with offset, which takes priority?  It's just redundant with offset in a nonstandard way IMO.  Not a blocking comment.

domino/routes.py
ddl-olsonJD: most of the previously written code is using v1/v4, Perhaps we should refactor all of them.

README.md
ddl-olsonJD: that makes sense.

README.md
ddl-olsonJD: No, I don't think so, as the admin may still want to apply other filters.

PR #214, opened 2025-02-20
PS-13958 Improve job management

hiiamelliott:
### Link to JIRA

https://onedominodatalab.atlassian.net/browse/PS-13958

python-domino had no function to restart failed jobs.
In addition, although it had a job_status function, there was no way to list previous jobs, so you had to know its ID already.

### What is the solution?

Adding the ability to list jobs using the `/jobs` API endpoint and the ability to restart them using the `/jobs/restart` endpoint.

### Testing

Tested by writing a script that uses the new `jobs_list()` function to get recent jobs, get any that had `statuses.executionStatus: Failed`, and then pass them back to the `job_restart()` function.
The test wasn't very intelligent– it didn't account for _why_ the jobs had failed– but you could also filter on `runFailureReason`/`runFailureMessage` to only restart jobs that had failed for a particular reason.

- [ ] Unit test(s)

### Pull Request Reminders

- [*] Has the [changelog](https://github.com/dominodatalab/python-domino/blob/master/CHANGELOG.md) been updated
- [*] Has relevant documentation been updated?
- [*] Does the code follow [Python Style Guide] (https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html)
- [*] Has the JIRA ticket(s) been linked above?


ddl-s-rodriguez (approved):
LGTM 🚀

PR #227, opened 2025-09-09
DOM-70889 DOM-70905  remove start time default and fix bug with filtering by run_id

niole:
The search_traces API defaults to using a filter start time of 1 hour ago. This was intended to limit the amount of data pulled into memory when there are many traces.

However, this API is now paginated and we don’t let the user get all traces at once. This default is not necessary. It is also confusing, because users may try to retrieve traces that were created over an hour ago by not specifying a start time. this was the case in a recent bug bash.

This PR also fixes a bug where the run_id wasn't being used correctly as a filter. traces from all runs in the experiment would appear in the response in stead of just the specified run.

### Link to JIRA

https://dominodatalab.atlassian.net/browse/DOM-70889
https://dominodatalab.atlassian.net/browse/DOM-70905

_placeholder_

### What is the solution?

_placeholder_

### Testing

- [ ] Unit test(s)

### Pull Request Reminders

- [ ] Has the [changelog](https://github.com/dominodatalab/python-domino/blob/master/CHANGELOG.md) been updated
- [ ] Has the JIRA ticket(s) been linked above?

### DIFF
--- a/domino/aisystems/tracing/tracing.py
+++ b/domino/aisystems/tracing/tracing.py
@@ -300,7 +300,8 @@ def search_traces(
 
     # get traces made within the time range
     # get traces with the trace names
-    filter_clauses = [time_range_filter_clause, trace_name_filter_clause]
+    run_filter_clause = f'metadata.mlflow.sourceRun = "{run_id}"'
+    filter_clauses = [run_filter_clause, time_range_filter_clause, trace_name_filter_clause]
     filter_string = ' AND '.join([x for x in filter_clauses if x])
 
     traces = client.search_traces(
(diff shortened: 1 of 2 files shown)


domino/aisystems/tracing/tracing.py
```diff
@@ -285,17 +285,23 @@ def search_traces(
...
+        time_range_filter_clause += f' AND timestamp_ms < {end_ms}'
 
     trace_name_filter_clause = None
     if trace_name:
         trace_name_filter_clause = f'trace.name = "{trace_name}"'
 
     # get traces made within the time range
     # get traces with the trace names
-    filter_clauses = [time_range_filter_clause, trace_name_filter_clause]
+    run_filter_clause = f'metadata.mlflow.sourceRun = "{run_id}"'
```
ddl-roshikiri, line 303: is this something we want to default as `None` for any reason or will we always want a run filter?

niole: we will always have it as part of the filter, because `run_id` is not optional in the arguments of the `search_traces` function in our sdk

PR #258, opened 2026-04-20
Fixes NameError when calling app_publish

domino-blake:

Calling app_publish()
line 958 where `payload` is referenced instead of `request`, causing a NameError. 

### What is the solution?

Update L958 to `request`


ddl-s-rodriguez (approved):

hutch3232:
@domino-blake @ddl-s-rodriguez I installed the latest version from pypi and now I'm hitting this bug. It seems it was merged with master branch but no bug fix release was ever made to pypi. Could we please fix this?

Issue #31, opened 2018-01-18
Blob Get By File Name


andscoop:
My current understanding of the API is that in order to download a file (blobs_get) you must have a file `key`.  In order to obtain a file `key` you must call `files_list` which requires `commitId`. In order to get a `commitId` you must run a file and check the status of that run.

Is there a better way to obtain the file key through the API? Currently downloading a file through the API is nearly pointless if I have to run a file first.

InbarRose:
I am having the same issue. I also want this solution.

It would be great to be able to download a file by simply providing the file-path (perhaps a different function). Even if I still have to provide the `commitId`. But I would also like to be able to leave `commitId` `None` and it simply take the `commitId` of the latest state of the project.

For anyone interested, here is the workaround I wrote using the logic you mentioned above:

```
def download_file(domino, file_path, local_path):
    # == start run : this can take a while ==
    # start a new run with a script that doesn't do anything
    run_results = domino.runs_start_blocking(command=['no-op-script.py'])
    # extract the run id
    run_id = run_results['runId']
    # fetch information about the run with the run_id
    run_info = domino.get_run_info(run_id)
    # find the commit_id of that run
    commit_id = run_info['commitId']  # maybe need to use 'outputCommitId'
    
    # == find file blob : this could fail, make sure you have the right path! ==
    # fetch all the files in this commit
    files_list = domino.files_list(commit_id)
    # make a list of files
    files = files_list['data']
    # iterate through all the files looking for file with matching path
    for f in files:
        if f['path'] == file_path:
            # save file key when we find the right file
            file_key = f['key']
            # stop looping over files
            break
    else:
        # if we didn't find the file, lets raise an exception
        raise RuntimeError('could not find file with path ({}) in commitId ({})'.format(file_path, commit_id))
    
    # == download file : saves the file to a local path ==
    # get the file object with the file key
    file_object = domino.blobs_get(file_key)
    # read the file object into a string
    file_content = file_object.read()
    # open a file locally to save file contents
    if not os.path.exists(os.path.dirname(local_path)):
        # if file save path dir does not exist we need to create it
        os.makedirs(os.path.dirname(local_path))
        # write file
        with open(local_path, 'wb') as f:
            f.write(file_content)
```

domino-blake:
Addressed by files_download() in #266

PR #266, opened 2026-04-22
new API additions — branch param, app_id, files_download, bug fixes

### LINKED ISSUE #24: Make a method to download a file without having to get the blob id first

### LINKED ISSUE #122: Document The Differnce Between domino.runs_start() and domino.job_start()
Both seem to be largely the same, but there do seem to be subtle differences between the two methods. However, it's not entirely clear what they are. It would be great if the documentation could identify what those differences are and when to use each method.

### LINKED ISSUE #127: Can we make Domino._app_id a public API?
Hi,

`Domino._app_id` is now 'protected', could it be a public API? We are making a domino product that needs to generate a URL to a deployed app, and for that, we need the app_id.
Happy to make a PR for this.

Regards,

Maarten Breddels

### LINKED ISSUE #128: Can we make Domino._Domino__app_get_status(app_id) a public API
Simar to #127, but for the app status.

Regards,

Maarten Breddels

### LINKED ISSUE #174: BUG: in _validate_hardware_tier_id
BUG!!!

This method is being passed a dict in the parameter hardware_tier_id, but the equality test expects a string!

1152     if hardware_tier_id == hardware_tier["hardwareTier"]["id"]:
   1153         return True
-> 1154 raise exceptions.HardwareTierNotFoundException(
   1155     f"{hardware_tier_id} hardware tier Id not found"
   1156 )

HardwareTierNotFoundException: {'value': 'small-k8s'} hardware tier Id not found

### LINKED ISSUE #231: Request: Allow job_start to specify the job's git repo branch to run on
Hi, thanks for this great adapter to the API. Would it be possible for the job_start method to receive besides the commit id, the branch to use for the job? Thanks.

domino-blake:
### What does this PR do?

Adds new backwards-compatible methods and parameters, plus one bug fix to `_validate_hardware_tier_id`. The only existing behaviour change is intentional: `_validate_hardware_tier_id` now accepts dict input from `compute_cluster_properties` instead of raising `HardwareTierNotFoundException` for a valid tier (#174).

A small number of `app_publish` / `app_unpublish` deprecation shims (commitId/appId) also land here because they're coupled to the new branch parameter see inline reply for context. Broader rename work lives in PR #267.

**New features**
- `job_start(branch=...)` — launch a job from a named branch without constructing a
  `mainRepoGitRef` dict manually. Mutually exclusive with `commit_id`. Closes #231.
- `app_id` property — previously private (`_app_id`), now a public read-only property
  returning the current project's app ID. Closes #127.
- `app_get_status(app_id)` — previously private (`__app_get_status`), now publicly
  accessible. Closes #128.
- `files_download(path, commit_id=None)` — convenience wrapper over `blobs_get_v2`
  that defaults to the latest commit, eliminating the need to call `commits_list()`
  manually. Closes #24, #31.

**Bug fix**
- `_validate_hardware_tier_id` now handles dict input (e.g. `{"value": "small-k8s"}`)
  from `compute_cluster_properties`, which previously caused `HardwareTierNotFoundException`
  even when the tier existed. Closes #174.

**Docs**
- Adds `runs_start` vs `job_start` comparison table to README.md and README.adoc. Closes #122.
- Documents all new methods and the `branch` param in both READMEs.

### Issues closed
#24, #31, #122, #127, #128, #174, #231

### Testing
- New tests added for every change:
  - `tests/test_jobs.py` — branch param, hardware tier dict input
  - `tests/test_app.py` — `app_get_status`
  - `tests/test_apps.py` — `app_id` property
  - `tests/test_files.py` — `files_download`

- [x] Unit tests passing
- [x] All changes are backwards compatible
- [x] Documentation updated

### DIFF
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -345,6 +345,7 @@ class Domino:
         external_volume_mounts: Optional[List[str]] = None,
         title: Optional[str] = None,
         main_repo_git_ref: Optional[dict] = None,
+        branch: Optional[str] = None,
     ) -> dict:
         """
         Starts a Domino Job via V4 API
@@ -415,9 +416,24 @@ class Domino:
                                                         "type": "branches",
                                                         "value": "my-feature-branch"
                                                     }
-                                                    Supported types: "branches", "tags"
+                                                    Supported types: "branches", "tags".
+                                                    Cannot be combined with branch.
+        :param branch:                              string (Optional)
+                                                    Convenience parameter. For git-based projects, launch the job
+                                                    from the tip of the specified branch. Cannot be combined with
+                                                    commit_id or main_repo_git_ref.
         :return: Returns created Job details (number, id etc)
         """
+        if branch and commit_id:
+            raise ValueError(
+                "Only one of branch or commit_id may be specified, not both."
+            )
+        if branch and main_repo_git_ref:
+            raise ValueError(
+                "Only one of branch or main_repo_git_ref may be specified, not both."
+            )
+        if branch:
+            main_repo_git_ref = {"type": "branches", "value": branch}
 
         def validate_on_demand_spark_cluster_properties(max_execution_slot_per_user):
             self.log.debug(
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -136,6 +136,10 @@ class Domino:
         publish_api_endpoint=None,
         **kwargs,
     ):
+        """
+        Start a run via the legacy v1 Runs API. For new work, prefer job_start() which uses
+        the v4 Jobs API and supports compute clusters, external volumes, and branch targeting.
+        """
         if "isDirect" in kwargs:
             warnings.warn(
                 "isDirect is deprecated, use is_direct",
@@ -427,7 +431,9 @@ class Domino:
         branch: Optional[str] = None,
     ) -> dict:
         """
-        Starts a Domino Job via V4 API
+        Start a Domino Job via the v4 Jobs API. Preferred over runs_start() for all new work —
+        supports hardware tiers, compute clusters (Spark/Ray/Dask/MPI), external volumes, and
+        git ref targeting by branch or commit.
         :param command:                             string
                                                     Command to execute in Job
                                                     >> domino.job_start(command="main.py arg1 arg2")
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -5,7 +5,7 @@ import os
 import re
 import time
 import warnings
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import polling2
 import requests
@@ -2104,7 +2104,9 @@ class Domino:
             f"{environment_id} environment not found"
         )
 
-    def _validate_hardware_tier_id(self, hardware_tier_id: str) -> bool:
+    def _validate_hardware_tier_id(self, hardware_tier_id: Union[str, Dict]) -> bool:
+        if isinstance(hardware_tier_id, dict):
+            hardware_tier_id = hardware_tier_id.get("value", hardware_tier_id)
         self.log.debug(f"Validating hardware tier id: {hardware_tier_id}")
         for hardware_tier in self.hardware_tiers_list():
             if hardware_tier_id == hardware_tier["hardwareTier"]["id"]:
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -891,6 +891,18 @@ class Domino:
         url = self._routes.blobs_get_v2(path, commit_id, project_id)
         return self.request_manager.get_raw(url)
 
+    def files_download(self, path: str, commit_id: Optional[str] = None):
+        """
+        Download a file from the project by path.
+
+        :param path: Path to the file within the project (e.g. "/README.md").
+        :param commit_id: The commit to download from. Defaults to the latest commit.
+        :return: Raw file content (urllib3 response stream).
+        """
+        if commit_id is None:
+            commit_id = self.commits_list()[0]["id"]
+        return self.blobs_get_v2(path, commit_id, self.project_id)
+
     def fork_project(self, target_name):
         url = self._routes.fork_project(self.project_id)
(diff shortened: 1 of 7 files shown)


domino/domino.py
```diff
@@ -610,10 +711,8 @@ def validate_is_external_volume_mounts_supported():
                 "masterHardwareTierId": master_hardware_tier_id,
             }
 
-        resolved_hardware_tier_id = hardware_tier_id or (
-            self.get_hardware_tier_id_from_name(hardware_tier_name)
-            if hardware_tier_name
-            else None
+        resolved_hardware_tier_id = (
```
ddl-bira-ignacio: issue: Looks like this now calls `get_hardware_tier_id_from_name` even when no tier name was passed. I think we need to keep this conditional

domino/domino.py
```diff
@@ -1533,7 +1711,7 @@ def hardware_tiers_list(self):
         url = self._routes.hardware_tiers_list(self.project_id)
         return self._get(url)
 
-    def get_hardware_tier_id_from_name(self, hardware_tier_name: str):
+    def get_hardware_tier_id_from_name(self, hardware_tier_name: Optional[str]):
```
ddl-bira-ignacio: suggestion: given `hardware_tier_name` is used here, I don't think it should be made optional.

ddl-bira-ignacio:
The description says "No existing behaviour changes" but there seems to be a few functional changes

domino/domino.py
ddl-bira-ignacio, line 1: issue: I see deprecation shims going in across a bunch of methods here, but the tests in this PR are focused on the new features. Would be great to either add shim-specific tests here or split that work into its own PR so coverage stays tight.

domino/domino.py
```diff
@@ -2069,20 +2249,15 @@ def project_id(self):
             f"Project '{self._project_name}' not found for owner '{self._owner_username}'"
         )
 
-    # This will fetch app_id of app in current project
     @property
-    def _app_id(self):
+    def app_id(self) -> Optional[str]:
```
ddl-bira-ignacio, line 2265: issue: `_app_id` has been around for a while and some folks might be using it even though it's private. Would it make sense to keep _app_id as a deprecated alias that points to app_id, at least for one release?

domino/domino.py
```diff
@@ -1944,7 +2122,9 @@ def _validate_environment_id(self, environment_id) -> bool:
             f"{environment_id} environment not found"
         )
 
-    def _validate_hardware_tier_id(self, hardware_tier_id: str) -> bool:
+    def _validate_hardware_tier_id(self, hardware_tier_id: Union[str, Dict]) -> bool:
+        if isinstance(hardware_tier_id, dict):
+            hardware_tier_id = hardware_tier_id.get("value", hardware_tier_id)
```
ddl-bira-ignacio: question: If the dict doesn't have a "value" key, `.get` just returns the original dict and the tier comparison will never match. Should we raise a clear error here instead of letting it fall through to "not found"?

domino/domino.py
```diff
@@ -791,6 +897,18 @@ def blobs_get_v2(self, path, commit_id, project_id):
...
+    def files_download(self, path: str, commit_id: Optional[str] = None):
+        """
+        Download a file from the project by path.
+
+        :param path: Path to the file within the project (e.g. "/README.md").
+        :param commit_id: The commit to download from. Defaults to the latest commit.
+        :return: Raw file content (urllib3 response stream).
+        """
+        if commit_id is None:
+            commit_id = self.commits_list()[0]["id"]
```
ddl-bira-ignacio: suggestion: Add a guard to `commits_list` in case it returns empty

ddl-bira-ignacio (commented):
The description says "No existing behaviour changes", but the diff also includes param renames and deprecation shims.

domino/domino.py
domino-blake: Fixed in eadc0a6. The conditional is restored — `get_hardware_tier_id_from_name` only runs when `hardware_tier_name` is non-None. Combined with reverting the parameter type to str 4140ce3, the call site and signature are now consistent.

domino/domino.py
domino-blake: Reverted to hardware_tier_name: str. With the call-site conditional restored in the previous commit, this function is never reached with None, so the original signature is correct.

domino/domino.py
domino-blake: Some shim work did land here alongside the new features rather than living cleanly in PR #267:
- app_publish / app_unpublish shims (commitId → commit_id, appId → app_id): these are coupled to PR B's new branch parameter on app_publish. Adding branch while leaving the adjacent params in camelCase would have shipped a mixed-casing API, so the renames came with the feature.
- runs_start, run_stop, runs_status, get_run_log, runs_stdout, files_list, endpoint_publish shims: you're right that these are unrelated to PR B's scope and leaked in via a "cherry pick" commit that bundled too much. They're also present on PR #267 (which is the rename PR), so they'll be deduplicated cleanly when PR #267 merges after this one.

If you'd prefer, I can either:

1. Strip the unrelated shims from this PR (would require rewriting history on commit 04e72a7 and the downstream commits that depend on it), or
2. Add shim coverage to this PR's test file for the ones that touch PR B's APIs (app_publish / app_unpublish), and let PR #267 cover the rest.

domino/domino.py
domino-blake: Added _app_id as a deprecated alias property that issues a DeprecationWarning and forwards to app_id. Existing callers will keep working for at least one release. 016ab96

domino/domino.py
domino-blake: Now raises a ValueError: dff1872

domino/domino.py
domino-blake: Added. commits_list() now gets stored, checked for empty, and a ValueError is raised b4b90d9

domino/domino.py
ddl-bira-ignacio: we could add more test coverage later

ddl-bira-ignacio (approved):

PR #265, opened 2026-04-22
linting, type safety, and CI pipeline

domino-blake:
### What does this PR do?

Adds automated quality gates and cleans up all pre-existing violations — no user-facing API changes.

**CI / tooling**
- Adds `.github/workflows/ci.yml` with three gating jobs: `lint` (black, isort, flake8,
  snake_case check), `typecheck` (mypy), and `test` (pytest matrix across Python 3.10/3.11/3.12)
- Adds `scripts/check_snake_case.py` — an AST-based checker that prevents new camelCase
  public parameter names from being introduced
- Rewrites `tox.ini` to cover all supported Python versions
- Updates all pre-commit hook versions to current releases
- Adds `pyproject.toml` with isort/black config

**Type safety**
- Resolves all 38 pre-existing mypy errors across 9 files (`domino/domino.py`,
  `domino/datasets.py`, `domino/_custom_metrics.py`, `domino/airflow/_operator.py`,
  `domino/agents/`)

**Formatting**
- Resolves all flake8, isort, and black violations across the codebase

### Testing
- 188 unit tests pass, 55 skipped (require live Domino deployment)
- All pre-commit checks pass: black, isort, flake8, check_snake_case, mypy

- [x] Unit tests passing
- [x] No user-facing API changes
- [x] CI jobs validated locally

---

PR #267, opened 2026-04-22
rename camelCase params to snake_case with deprecation shims

domino-blake:
### What does this PR do?

Renames 18 public API parameters from camelCase to snake_case to comply with PEP 8.
Old names continue to work and emit a `DeprecationWarning` pointing callers to the
new name — no hard breaking changes in this PR.

**Renamed parameters**

| Method | Old name | New name |
|--------|----------|----------|
| `runs_start` / `runs_start_blocking` | `isDirect` | `is_direct` |
| `runs_start` / `runs_start_blocking` | `commitId` | `commit_id` |
| `runs_start` / `runs_start_blocking` | `publishApiEndpoint` | `publish_api_endpoint` |
| `run_stop` | `saveChanges` | `save_changes` |
| `runs_status` | `runId` | `run_id` |
| `get_run_log` | `includeSetupLog` | `include_setup_log` |
| `runs_stdout` | `runId` | `run_id` |
| `files_list` | `commitId` | `commit_id` |
| `endpoint_publish` | `modelId` / `modelVersionId` | `model_id` / `model_version_id` |
| `app_publish` | `unpublishRunningApps` / `hardwareTierId` / `environmentId` / `externalVolumeMountIds` / `commitId` / `appId` | snake_case equivalents |

**Tests**
- Adds `tests/test_deprecations.py` with full coverage: every old name raises
  `DeprecationWarning`, every new name is accepted silently.

### Note for reviewers
These are technically breaking in a future major version (old names will eventually
be removed). The deprecation shims give callers time to migrate. Happy to discuss
the timeline for removal if needed.

### Testing
- 18 new tests in `tests/test_deprecations.py`
- All existing tests continue to pass (they use the old names, which still work)

- [x] Unit tests passing
- [x] Deprecation warnings added for all renamed params
- [x] CHANGELOG updated
- [x] README updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,12 @@ jobs:
       - name: flake8
         run: flake8 .
 
+      - name: snake_case
+        run: |
+          find domino -name "*.py" \
+            | grep -v "domino/_impl/" \
+            | grep -v "domino/airflow/" \
+            | xargs python scripts/check_snake_case.py
 
   typecheck:
     name: Type check
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,11 +30,10 @@ jobs:
         run: flake8 .
 
       - name: snake_case
-        run: |
-          find domino -name "*.py" \
-            | grep -v "domino/_impl/" \
-            | grep -v "domino/airflow/" \
-            | xargs python scripts/check_snake_case.py
+        # check_snake_case.py owns the include/exclude rules; pass it everything
+        # under domino/ and it will skip paths that don't apply. Keeps CI in
+        # lockstep with the pre-commit hook (both invoke the same script).
+        run: find domino -name "*.py" | xargs python scripts/check_snake_case.py
 
   typecheck:
     name: Type check
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -38,6 +38,7 @@ from domino.http_request_manager import _HttpRequestManager
 from domino.routes import _Routes
 
 
+
 class Domino:
     def __init__(
         self,
--- a/domino/domino.py
+++ b/domino/domino.py
@@ -37,6 +37,35 @@ from domino.domino_enums import (
 from domino.http_request_manager import _HttpRequestManager
 from domino.routes import _Routes
 
+# Sentinel used as the default value of renamed parameters so we can detect
+# whether the caller passed the new name, the deprecated name, or neither.
+_UNSET: Any = object()
+
+
+def _resolve_renamed_kwarg(new_value, old_name, new_name, kwargs, default):
+    """Resolve a renamed parameter, raising if both old and new names were passed.
+
+    If the deprecated name is present in kwargs, emit a DeprecationWarning and
+    return its value. If both the new name (non-sentinel) and the deprecated
+    name are provided in the same call, raise ValueError so the caller can fix
+    the ambiguity instead of silently picking one.
+    """
+    has_old = old_name in kwargs
+    has_new = new_value is not _UNSET
+    if has_old and has_new:
+        raise ValueError(
+            f"Pass either '{new_name}' or '{old_name}', not both. "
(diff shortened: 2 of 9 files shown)


domino-blake:
Once #265 merges, need to change this to target master

ddl-bira-ignacio:
issue: The table says endpoint_publish renames modelId/modelVersionId, but the actual diff renames commitId to commit_id. Looks like the table just needs a small fix.

domino/domino.py
```diff
@@ -129,22 +129,44 @@ def runs_list(self):
...
+        publish_api_endpoint=None,
+        **kwargs,
     ):
+        if "isDirect" in kwargs:
+            warnings.warn(
+                "isDirect is deprecated, use is_direct",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            is_direct = kwargs.pop("isDirect")
```
ddl-bira-ignacio: issue: If someone passes both is_direct=... and isDirect=... in the same call, the old name silently wins here because `kwargs` gets popped after binding. Might be worth raising a `ValueError` when both are provided so nobody gets surprised.

domino/domino.py
```diff
@@ -129,22 +129,44 @@ def runs_list(self):
...
         command,
-        isDirect=False,
-        commitId=None,
+        is_direct=False,
+        commit_id=None,
         title=None,
         tier=None,
-        publishApiEndpoint=None,
+        publish_api_endpoint=None,
+        **kwargs,
```
ddl-bira-ignacio, line 166: question: What's the idea behind adding kwargs here? it looks like the method only uses well known names. We should probably adopt that instead. 
if it's just to deal with parameter naming, then i'd suggest putting a deprecation warning and do a full replace in the next release

.github/workflows/ci.yml
```diff
@@ -29,6 +29,12 @@ jobs:
       - name: flake8
         run: flake8 .
 
+      - name: snake_case
+        run: |
+          find domino -name "*.py" \
```
ddl-bira-ignacio: suggestion: The find/grep combo here seems different than what the pre-commit hook does

domino/domino.py
ddl-bira-ignacio, line 1: note: Just want to call out that these renames are Python-side only -- the JSON keys going over the wire are still camelCase. Might be worth a quick note in the README or changelog so users don't think the API contract changed.

domino/domino.py
domino-blake: silent override is fixed. b45ea77

domino/domino.py
domino-blake: The **kwargs is purely the deprecation mechanism — every method that has it is using it to absorb the old camelCase name without breaking existing callers

.github/workflows/ci.yml
domino-blake: fixed by moving the filtering rules into check_snake_case.py itself, so both CI and pre-commit invoke the same 9987404

domino/domino.py
domino-blake: Added a note in CHANGELOG.md, README.md, and README.adoc explicitly calling out that these renames are Python-side only bd1785b

domino/domino.py
```diff
@@ -268,31 +326,44 @@ def runs_start_blocking(
 
         return run_response
 
-    def run_stop(self, runId, saveChanges=True):
+    def run_stop(self, run_id=_UNSET, save_changes=_UNSET, **kwargs):
```
ddl-bira-ignacio, line 333: suggestion: `run_stop` with no args now silently passes None as job_id instead of raising, as it would before. It might be worth putting code that checks if there's a valid run id passed in

domino-blake: Added a check on [9a24604](https://github.com/dominodatalab/python-domino/pull/267/commits/9a24604dfc11c8c1c5349f272a90b8bf1a964a48)

ddl-bira-ignacio (approved):

